PR #8, opened 2018-07-30
Add mission statement and other info to readme

rafalp:
This PR adds following to our readme:

- small mission statement
- warning about lib currently being experimental
- badges for our tests suite

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,8 @@
-# ariadne
-A functional implementation of a Python GraphQL server
+# Ariadne: Build functional GraphQL APIs with just enough amount of magic.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,8 @@
-# ariadne
-A functional implementation of a Python GraphQL server
+# Ariadne: Build functional GraphQL APIs with just enough amount of magic.
```
patrys: Suggestion: drop "amount of" and the trailing dot

maarcingebala: Maybe: "Build functional GraphQL APIs in Python [...]"? If we want to build hype around it in the open-source community we should make it clear and obvious from the first sentence.

PR #10, opened 2018-08-01
Change library structure

rafalp:
This PR moves code around so it should be easier to work on individual features in isolation.

It also includes crude quickstart guide to readme illustrating how to create basic executable schema, run an query on it, then access resulting data.


README.md
```diff
@@ -5,4 +5,65 @@
...
+        query Query
+    }
+
+    type Query {
+        persons [Person]!
+    }
+
+    type Person {
+        firstName String
+        secondName String
```
maarcingebala: Here we have `secondName` field while in the resolver there is `lastName`. I think `lastName` fits better.

rafalp:
@maarcingebala very nice catch! I've fixed that together with few other issues that quickstart code contained.

README.md
patrys: I would say that l10n-wise the _correct_ way to do it is to have `givenName` and `familyName` 😉

README.md
```diff
@@ -5,4 +5,67 @@
...
-**Warning**: The work on library is currently in experimental phase. We invite you to give Ariadne a try, but if you are looking to build production-ready GraphQL API, please use more stable solutions such as [Graphene](https://github.com/graphql-python/graphene).
\ No newline at end of file
+**Warning**: The work on library is currently in experimental phase. We invite you to give Ariadne a try, but if you are looking to build production-ready GraphQL API, please use more stable solutions such as [Graphene](https://github.com/graphql-python/graphene).
+
+## Quickstart 
+
+```python
+from ariadne import execute_request, make_executable_schema
+
+
```
patrys: Extra line 😄

README.md
```diff
@@ -5,4 +5,67 @@
...
+type_defs = """
+    schema {
+        query: Query
+    }
+
+    type Query {
+        people: [Person]!
```
patrys: Shouldn't this be `[Person!]!`?

PR #15, opened 2018-08-10
Automatically define schema based on types

patrys:
Automatically define the schema based on existing types if one is not already provided.

Allows us to use examples closer to these used by Apollo.

Unfortunately I don't have the time necessary to provide a better test suite.


rafalp:
Very nice! I'll follow up with updated readme and extra tests in the future.

PR #16, opened 2018-08-20
Remove schema from readme example

rafalp:
Update quickstart in readme to dont show `schema {}`, because we don't need it anymore.


rafalp:
@patrys yup, the plan is to stress that those are type definitions, not actual schema.

PR #19, opened 2018-08-20
Add build_schema_from_type_definitions

### LINKED ISSUE #7: make_executable_schema type_defs arg should accept str or list of str
The `make_executable_schema` utility should take string or list of strings with SDLs for type definitions and dict of dicts with resolvers as second argument, parse each one, merge them, then, if resolvers are given, add them to schema returned by `build_ast_schema` util from `graphql-core`, and finally return resulting executable schema that may then be passed to `graphql()`.

rafalp:
This PR:

- renames `build_schema` to `build_schema_from_type_definitions` (after apollo)
- adds support for definitions concatenating
- adds some tests and moves code around
- cut module names length and drop "function per module" approach

Fixes #7 

## TODO

* [x] Test happy path
* [x] Test what happens if type is redefined

### DIFF
--- a/ariadne/make_executable_schema.py
+++ b/ariadne/make_executable_schema.py
@@ -1,11 +1,12 @@
+from typing import List, Union
+
 from graphql import GraphQLSchema
 
 from .add_resolve_functions_to_schema import add_resolve_functions_to_schema
 from .build_schema_from_type_definitions import build_schema_from_type_definitions
-from .types import TypeDefs
 
 
-def make_executable_schema(type_defs: TypeDefs, resolvers: dict) -> GraphQLSchema:
+def make_executable_schema(type_defs: Union[str, List[str]], resolvers: dict) -> GraphQLSchema:
     schema = build_schema_from_type_definitions(type_defs)
     add_resolve_functions_to_schema(schema, resolvers)
     return schema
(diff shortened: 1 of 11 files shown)


ariadne/make_executable_schema.py
```diff
@@ -1,54 +1,14 @@
...
-
-
-def build_schema(type_defs: str) -> GraphQLSchema:
-    document = parse(type_defs)
-    if not document_has_schema(document):
-        schema_definition = build_default_schema(document)
-        document.definitions.append(schema_definition)
-    return build_ast_schema(document)
+from .add_resolve_functions_to_schema import add_resolve_functions_to_schema
+from .build_schema_from_type_definitions import build_schema_from_type_definitions
```
patrys: These names are enormous :/

patrys: I would try to keep module names smaller even if you prefer to match the function names of Apollo.

rafalp: @patrys I've moved some code around, made our modules area based (eg. resolvers, build schema, executable schema) and renamed modules to reflect that, giving us shorter modules now:

`resolvers.py`
`build_sechma.py`
`make_executable_schema`

rafalp:
Added extra acceptance test checking if default resolver works with object types - bumping our coverage.

PR #28, opened 2018-09-14
Add GraphQL Playground

### LINKED ISSUE #9: GraphQL API explorer
It is a standard among other GraphQL frameworks to ship a built-in GraphQL API explorer. We should consider adding one in Ariadne as well:
- [GraphiQL](https://github.com/graphql/graphiql)
- [graphql-playground](https://github.com/prismagraphql/graphql-playground)

Probably we should choose the second one as it allows to set headers and contains queries history.

rafalp:
This PR adds `ariadne.wsgi` module that provides `GraphQLMiddleware` WSGI Middleware that displays Apollo Playground on `GET` and executes GraphQL query on `POST`.

In addition to this, `GraphQLMiddleware` also provides `make_simple_server` classmethod that reeturns `wsgiref.simple_server` using above middleware to allow people to do some quick hacking with Ariadne/GraphQL without need for full-fledged web framework.

Implementation is still super hacky, but its happening.

### TODO

- [x] Split introspection fix to separate PR (#29)
- [x] Query root
- [x] Query context
- [x] Validate that query's json deserializes to dict and not list
- [x] `GraphQLMiddleware`
- [x] Type annotations
- [x] Tests
- [x] Update example code in readme
- [x] Don't (ab)use fixtures by using it as function factories

Fixes #9


ariadne/wsgi.py
```diff
@@ -0,0 +1,177 @@
...
+from typing import List, Union
+from wsgiref.simple_server import make_server
+
+from graphql import format_error, graphql
+
+from .executable_schema import make_executable_schema
+
+JSON_CONTENT_TYPE = "application/json"
+
+PLAYGROUND_MINIMAL = """
```
salwator: Could we extract html from python module?

ariadne/wsgi.py
```diff
@@ -0,0 +1,183 @@
...
+        return None
+
+    def get_query_context(
+        self, environ, request_data: dict
+    ):  # pylint: disable=unused-argument
+        """Override this method in inheriting class to create query context."""
+        return {"environ": environ}
+
+    def return_response_from_result(self, start_response, result) -> List[str]:
+        status = "200 OK"
```
salwator: we should extract magic values

ariadne/wsgi.py
```diff
@@ -0,0 +1,183 @@
...
+
+        start_response(status, [("Content-Type", JSON_CONTENT_TYPE)])
+        return [json.dumps(response).encode("utf-8")]
+
+    @classmethod
+    def make_server(
+        cls, type_defs: Union[str, List[str]], resolvers: dict, port: int = 8888
+    ):
+        wsgi_app = cls(None, type_defs, resolvers)
+        return make_server("127.0.0.1", port, wsgi_app)
```
salwator: Shouldn't we inject IP?

ariadne/playground.py
```diff
@@ -0,0 +1,60 @@
...
+  </div>
+  <script>window.addEventListener('load', function (event) {
+      GraphQLPlayground.init(document.getElementById('root'), {
+        // options as 'endpoint' belong here
+      })
+    })</script>
+</body>
+
+</html>
+""".strip()
```
NyanKiyoshi: Why not just putting a backslash like this:
```python
"""\
Hello
world\
"""
```

Instead of striping just for two `\n`?

rafalp: Personal preference - I don't like the `\` and startup performance hit from `.strip()` is minimal. ;)

ariadne/wsgi_middleware.py
```diff
@@ -0,0 +1,160 @@
...
+
+        start_response(status, [("Content-Type", CONTENT_TYPE_JSON)])
+        return [json.dumps(response).encode("utf-8")]
+
+    @classmethod
+    def make_simple_server(
+        cls,
+        type_defs: Union[str, List[str]],
+        resolvers: Union[dict, List[dict]],
+        port: int = 8888,
```
NyanKiyoshi: Why not allowing to pass the host? 
Also, maybe we could add a little docstring with an example usage? e.g.:

```python
>>> type_defs = """
...    type Query {
...        test: String
...    }
... """
>>> host = "0.0.0.0"
>>> port = 8888
>>>
>>> playground_httpd = GraphQLMiddleware.make_simple_server(
...     type_defs=type_defs, resolvers={}, host=host, port=port)
...
>>> print(f'Now serving on http://{host}:{port}')
>>> playground_httpd.serve_forever()
>>> 
```

I know it's kinda temporary and hacky PR! Awesome work anyway :+1:

NyanKiyoshi: For a development server, I can't see a real difference unlike a WSGI production server.

Apart that I think it is odd that we are making the development server to listen on the local network interface (`0.0.0.0`) instead of the `loopback` interface. As this would expose the development application to the rest of the local network (unless the firewall is blocking it of course), I feel like it would be an unwanted behavior.

Maybe we should default it `127.0.0.1` and allow the user to set the listening host to `0.0.0.0` if the user explicitly wants to?

Or am I missing a reason that the server is being defaulted (or hardcoded) to `0.0.0.0`?

rafalp: Letting user specify the host, even if its `0.0.0.0` vs default `127.0.0.1` is good enough reason to make host configurable 👍

I'll change that, thanks!

ariadne/playground.py
NyanKiyoshi: @rafalp do you have the button to mark the conversation as resolved? I don't why but I can no longer mark review as resolved, not sure if github changed the mechanism or if it's a bug or something.

rafalp: AFAIK you can't resolve comments, only changes requested, because repo can be setup to block unapproved merges.

rafalp:
Idea for future: extract some of logic from WSGI Middleware to reusable functions, to make it easier for 3rd party lib users to implement custom GraphQL servers (we could potentially share some codebase in prospective Django GraphQL view.)

.gitignore
```diff
@@ -109,3 +109,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# demo
+demo.py
```
patrys: This is rather unusual.

mavaras:
Hi everyone, I have a question. Is there any way to customize the Playground HTTP headers through this implementation? Thanks !

rafalp:
@mavaras please ask on our [Spectrum](https://spectrum.chat/ariadne).

PR #29, opened 2018-09-14
Schema introspection

rafalp:
This PR changes the `add_resolve_functions_to_object` so it no longer breaks types that already had resolvers defined on them, namely `__Schema` type and friends.

### TODO

* [x] Test that schema returned by `make_executable_schema` can be introspected

### DIFF
--- a/ariadne/resolvers.py
+++ b/ariadne/resolvers.py
@@ -30,8 +30,11 @@ def add_resolve_functions_to_schema(schema: GraphQLSchema, resolvers: dict):
 def add_resolve_functions_to_object(name: str, obj: GraphQLObjectType, resolvers: dict):
     type_resolvers = resolvers.get(name, {})
     for field_name, field_object in obj.fields.items():
-        field_resolver = type_resolvers.get(field_name, default_resolver)
-        field_object.resolver = field_resolver
+        field_resolver = type_resolvers.get(field_name)
+        if field_resolver:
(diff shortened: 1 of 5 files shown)


ariadne/resolvers.py
```diff
@@ -30,8 +30,11 @@ def add_resolve_functions_to_schema(schema: GraphQLSchema, resolvers: dict):
 def add_resolve_functions_to_object(name: str, obj: GraphQLObjectType, resolvers: dict):
     type_resolvers = resolvers.get(name, {})
     for field_name, field_object in obj.fields.items():
-        field_resolver = type_resolvers.get(field_name, default_resolver)
-        field_object.resolver = field_resolver
+        field_resolver = type_resolvers.get(field_name)
+        if field_resolver:
```
salwator: Could be extracted to function for better readability.

tests/test_introspection.py
```diff
@@ -0,0 +1,26 @@
...
+        name: String
+        age: Int!
+        dateOfBirth: Date!
+    }
+
+    scalar Date
+"""
+
+
+def test_introspect_executable_schema():
```
salwator: I would rename test to express its restricted scope to no-exception check.
Other thing is, we already use tests like those to integrate *ariadne* with *graph-core*, so we test expected output that depends on external library. Maybe, not saying that definitely, we could treat this test the same and check expected introspection response.

PR #30, opened 2018-09-24
Accept list of resolver maps

### LINKED ISSUE #11: make_executable_schema resolvers arg should accept dict of dicts or list of dicts of dicts
The `make_executable_schema` utility should optionally take list of dicts of dicts (AKA "resolvers map"), this would allow larger projects to easily split and compose resolvers as needed:

```python
from ariadne import make_executable_schema

from products.graphql import resolvers as products_resolvers
from users.graphql import resolvers as users_resolvers


typedefs = "..."
resolvers = [products_resolvers, users_resolvers)

schema = make_executable_schema(typedefs, resolvers)
```

salwator:
Closes #11

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -218,3 +218,43 @@ def test_mapping_resolver_to_object_attribute():
     result = graphql(schema, "{ user { firstName } }")
     assert result.errors is None
     assert result.data == {"user": {"firstName": "Joe"}}
+
+
+def test_accept_list_of_resolvers_map():
+    type_defs = """
+        type Query {
+            user: User
+            test(data: TestInput): Int
+        }
+
+        type User {
+            firstName: String
+        }
+
+        input TestInput {
+            value: Int
+        }
+    """
+
+    def resolve_test(*_, data):
+        assert data == {"value": 4}
+        return "42"
+
+    resolvers = [{
+        "Query": {"user": lambda *_: Mock(first_name="Joe")},
+        "User": {"firstName": resolve_to("first_name")},
+    },
+        {
+            "Query": {"test": resolve_test}
+        }
+    ]
+
+    schema = make_executable_schema(type_defs, resolvers)
+
+    result = graphql(schema, "{ test(data: { value: 4 }) }")
+    assert result.errors is None
+    assert result.data == {"test": 42}
+
+    result = graphql(schema, "{ user { firstName } }")
+    assert result.errors is None
+    assert result.data == {"user": {"firstName": "Joe"}}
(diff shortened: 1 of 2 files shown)


tests/test_queries.py
```diff
@@ -218,3 +218,42 @@ def test_mapping_resolver_to_object_attribute():
     result = graphql(schema, "{ user { firstName } }")
     assert result.errors is None
     assert result.data == {"user": {"firstName": "Joe"}}
+
+
+def test_accept_list_of_resolvers_map():
```
rafalp: Shouldn't this be `resolvers_maps`?

tests/test_queries.py
```diff
@@ -218,3 +218,42 @@ def test_mapping_resolver_to_object_attribute():
...
+    def resolve_test(*_, data):
+        assert data == {"value": 4}
+        return "42"
+
+    resolvers = [
+        {
+            "Query": {"user": lambda *_: Mock(first_name="Joe")},
+            "User": {"firstName": resolve_to("first_name")},
+        },
+        {"Query": {"test": resolve_test}},
```
rafalp: What happens when later resolver map redefines the resolver defined in previous one? We'll need to know this and have tests harness in place (just like how we are testing merging typedefs in `test_modularization`).

For reference, here's [`lodash.merge`](https://lodash.com/docs/4.17.10#merge) mechanism used by Apollo-Server:

> It recursively merges own and inherited enumerable string keyed properties of source objects into the destination object. Source properties that resolve to undefined are skipped if a destination value exists. Array and plain object properties are merged recursively. Other objects and value types are overridden by assignment. Source objects are applied from left to right. Subsequent sources overwrite property assignments of previous sources.

rafalp:
Nice, with this we'll wrap up the support for schema modularization, one of must-have features for api devs.

I have doubts about where this code lives now. Its part of larger feature (let's call it "API code modularization") but:

- We have `test_modularization` that contains tests for modularizing the typedefs. Yet test for modularizing resolvers lives in `test_queries`. Considering that use case/scenario is same, I would look for those in same place.
- Code for merging type defs lives in `build_schema`, code for adding resolvers, default/utility resolvers in `resolvers`, and code for merging resolvers map lives in `executable_schema`. This means we now have two different modules for processing resolvers maps.

Feng shui feels off to me here. Perhaps `add_resolve_functions_*` should be moved to executable schema? Or maybe we should have dedicated module for merging typedefs?

salwator:
@rafalp I have reverted type hints, we could reintroduce them again after dropping 3.5 one day.

PR #35, opened 2018-10-04
If the attribute is callable, call it with query arguments

### LINKED ISSUE #25: If value from resolver is callable, call it.
[Apollo doc](https://www.apollographql.com/docs/graphql-tools/resolvers) for default resolver says that if `field_name` resolves to function, it will be called with query arguments:

> Calls a function on obj with the relevant field name and passes the query arguments into that function

This can be useful for situations when parent resolver returned an object with getter functions.

bogdal:
Fixes #25

### DIFF
--- a/ariadne/resolvers.py
+++ b/ariadne/resolvers.py
@@ -2,10 +2,13 @@ from graphql import GraphQLObjectType, GraphQLScalarType, GraphQLSchema
 from graphql.execution.base import ResolveInfo
 
 
-def resolve_parent_field(parent, name: str):
+def resolve_parent_field(parent, name: str, **args: dict):
     if isinstance(parent, dict):
         return parent.get(name)
-    return getattr(parent, name, None)
+    attribute = getattr(parent, name, None)
+    if callable(attribute):
+        return attribute(**args)
+    return attribute
 
 
 def default_resolver(parent, info: ResolveInfo):
@@ -13,8 +16,8 @@ def default_resolver(parent, info: ResolveInfo):
 
 
 def resolve_to(name: str):
-    def resolver(parent, *_):
-        return resolve_parent_field(parent, name)
+    def resolver(parent, *_, **args):
+        return resolve_parent_field(parent, name, **args)
 
     return resolver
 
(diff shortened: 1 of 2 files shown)


ariadne/resolvers.py
```diff
@@ -2,19 +2,22 @@
 from graphql.execution.base import ResolveInfo
 
 
-def resolve_parent_field(parent, name: str):
+def resolve_parent_field(parent, name: str, **args: dict):
```
rafalp: `**args` should be renamed to `**kwargs`

ariadne/resolvers.py
```diff
@@ -2,19 +2,22 @@
...
     if isinstance(parent, dict):
         return parent.get(name)
```
rafalp: What if dict's value is callable too?

ariadne/resolvers.py
```diff
@@ -2,19 +2,22 @@
...
-    return getattr(parent, name, None)
+    attribute = getattr(parent, name, None)
```
rafalp: How about `value`?

tests/test_queries.py
```diff
@@ -196,3 +196,36 @@ def test_mapping_resolver_to_object_attribute():
     result = graphql(schema, "{ user { firstName } }")
     assert result.errors is None
     assert result.data == {"user": {"firstName": "Joe"}}
+
+
+def test_mapping_resolver_to_object_function():
```
rafalp: @salwator has converted me to story-telling test names, and we would like to move our tests suite to those. Perhaps `test_default_resolver`

I'm also thinking that test scenario could be more intuitive/real-worldly. Maybe we could have `logs: Logs` resolver accepting `env: String!`, asserting that it has valid value and returning Mock with `count` resolving to `get_items_count()`?

ariadne/resolvers.py
bogdal: Good point 👍

tests/test_queries.py
```diff
@@ -196,3 +196,58 @@ def test_mapping_resolver_to_object_attribute():
...
+        blogPosts(published: $published)
+      }
+    }
+    """
+    variables = {"size": "200x300", "published": True}
+
+    result = graphql(schema, query, variables=variables)
+    assert result.errors is None
+    assert result.data == {
+        "user": {"firstName": "Joe", "avatar": "test-url", "blogPosts": 3}
```
salwator: We might want to reuse those constants in mock and assertion so it would be enough to change one number in future.

Issue #21, opened 2018-08-22
Create `resolve_to` utility


rafalp:
Python variables are usually named using the `snake_case`, whereas majority of JS written today uses the `camelCase`. Ariadne will need to provide a way to resolve one to another.

The idea is to create `resolver_for` function taking single `name` argument and returning resolver for specified attr/key name, enabling **explicit** mapping as such:

```
resolvers_map = {
    "User": {
        "lastVisitedOn": resolve_to("last_visited_on")
    }
}
```

Other advantage of having such utility would be arbitrary field names:

```
resolvers_map = {
    "User": {
        "lastVisitedOn": resolve_to("last_visit")
    }
}
```

PR #36, opened 2018-10-05
Add expose function

salwator:
Since issue #21 is solved, we can define resolvers as follow:

```python
resolvers = {
    "User":  {"firstName": resolve_to("first_name")},
}
```

It seems to be common boilerplate for future uses, since we will repeat this line for every field, when we need to connect python and GraphQL naming conventions.

In this PR I developed experiment *expose* util function to help create it easier:

```python
resolvers = {
   "User": expose("first_name", "last_name"),
}
```

that will be equivalent to
```python
resolvers = {
    "User":  {
        "firstName": resolve_to("first_name"),
        "lastName": resolve_to("last_name"),
    },
}
```

### DIFF
--- a/ariadne/__init__.py
+++ b/ariadne/__init__.py
@@ -1,6 +1,11 @@
 from .build_schema import build_schema_from_type_definitions
 from .executable_schema import make_executable_schema
-from .resolvers import add_resolve_functions_to_schema, default_resolver, resolve_to
+from .resolvers import (
+    add_resolve_functions_to_schema,
+    default_resolver,
+    expose,
+    resolve_to,
+)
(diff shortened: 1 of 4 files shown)


rafalp:
We've been discussing something like that internally ~1.5month ago, and general sentiment was that we don't want to have scaffolds like this or automatic case conversion in the lib. Short argument was `explicit is better than implicit`, but I can't remember what was longer argument brought from our Saleor experiences.

Marking this as "decision needed" for now, so we may revisit this later.

patrys:
Pretty please, do not make the tool automagically perform string translation.

PR #38, opened 2018-10-16
Introduce enums

### LINKED ISSUE #27: Implement support for Enums
Spec: https://graphql.org/learn/schema/#enumeration-types

salwator:
We need to be compatible with [Enum definitions](https://graphql.org/learn/schema/#enumeration-types).
This PR explores this functionality from testing side and ensures our compatibility for future.

fixes #27

### DIFF
--- a/tests/test_enums.py
+++ b/tests/test_enums.py
@@ -19,14 +19,14 @@ TEST_VALUE = "NEWHOPE"
 INVALID_VALUE = "LUKE"
 
 
-def test_succesfull_enum():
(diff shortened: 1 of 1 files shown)


rafalp:
I'm surprised that enums support requires 0 work inside the library to support ;)

In future releases we'll want to support conversion to internal form, [like how Apollo server does it](https://www.apollographql.com/docs/apollo-server/features/scalars-enums.html#internal-values), but that can wait for after 0.1

PR #41, opened 2018-10-17
Add basic documentation

rafalp:
This PR adds basic documentation to the library.

### TODO

- [x] Add homepage
- [x] Add introduction
- [x] Add resolvers (how to, default, resolve_to)
- [x] Mutations
- [x] Error messaging
- [x] Custom scalars
- [x] Enums
- [x] Modularization
- [x] WSGI middleware
- [x] Writing custom servers


docs/custom-server.rst
```diff
@@ -0,0 +1,132 @@
...
+
+In addition to simple GraphQL server implementation in form of ``GraphQLMiddleware``, Ariadne provides building blocks for assembling custom GraphQL servers.
+
+
+Creating executable schema
+--------------------------
+
+The key piece of GraphQL server is *executable schema* - an schema with resolver functions attached to fields.
+
+Ariadne provides ``make_executable_schema`` utility function that takes type definitions as first argument, and resolvers map as second, and returns executable instance of ``GraphQLSchema``::
```
patrys: Sphinx provides semantic macros to aid with documenting code that will automatically link to the definition if you provide one. Maybe we should make use of these and of autodoc to provide both a guide and an API reference?

```
Ariadne provides :py:func:`make_executable_schema` ...
```

rafalp: Noted, I've added those to some of our docs, just not the whole picture.

I think for 0.1 we'll stick to what we have already, and then improve incrementally in future releases.

PR #43, opened 2018-10-25
Rename resolve_to to resolve

rafalp:
This PR renames `resolve_to` utility to just `resolve`, based on feedback we've got yesterday on internal demo we've did @mirumee.

### DIFF
--- a/ariadne/__init__.py
+++ b/ariadne/__init__.py
@@ -1,6 +1,6 @@
 from .build_schema import build_schema_from_type_definitions
 from .executable_schema import make_executable_schema
-from .resolvers import add_resolve_functions_to_schema, default_resolver, resolve_to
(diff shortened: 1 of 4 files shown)


patrys:
```python
from functools import wraps
import inspect

def simple_resolver(func):
    @wraps(func)
    def inner(root, info, **kwargs):
        return func(root, **kwargs)

    return inner

def wrap_resolver(resolver):
    signature = inspect.signature(resolver)
    if 'info' not in signature.parameters:
        return simple_resolver(resolver)
    return resolver
```

...

```python
from operator import attrgetter

resolvers = {
    "Foo": {
        "field_alias": attrgetter("field_name")
    }
}
```

rafalp:
Punting on this PR, I'll be making roadmap item for resolvers redo based on what @patrys proposed later..

PR #58, opened 2018-11-12
Define object types for mutations

bogdal:
This PR defines custom object types for exemplary mutations. It's a good practice to define an object type for each migration separately, even if they initially contain the same fields.

### DIFF
--- a/docs/mutations.rst
+++ b/docs/mutations.rst
@@ -33,7 +33,7 @@ In this example we have the following elements:
 
 ``Mutation`` type with two mutations: ``login`` mutation that requires username and password strings and returns bool with status, and ``logout`` that takes no arguments and just returns status.
 
-For the sake of simplicity, our mutations return bools, but really there is no such restriction. You can have a resolver that returns status code, an updated object, or an error message::
+For the sake of simplicity, our mutations return bools, but really there is no such restriction. In fact, it is recommended for mutations to return dedicated ``Payload`` types:
 
     type_def = """
         type Mutation {
@@ -47,6 +47,7 @@ For the sake of simplicity, our mutations return bools, but really there is no s
         }
     """
 
+Above mutation will return special type containing information about mutation's status, as well as either ``Error`` message or logged in ``User``.
 
 Writing resolvers
 -----------------
@@ -99,10 +100,10 @@ Imagine a mutation for creating ``Discussion`` that takes category, poster, titl
 
     type_def = """
         type Mutation {
-            createDiscussion(category: ID!, title: String!, isAnnouncement: Boolean, isClosed: Boolean): CreateDiscussionPayload
+            createDiscussion(category: ID!, title: String!, isAnnouncement: Boolean, isClosed: Boolean): DiscussionPayload
         }
 
-        type CreateDiscussionPayload {
+        type DiscussionPayload {
             status: Boolean!
             error: Error
             discussion: Discussion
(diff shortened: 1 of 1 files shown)


bogdal:
Sure, I agree that the examples should be as simple as possible, as long as the schema is valid.

docs/mutations.rst
```diff
@@ -33,18 +33,21 @@ In this example we have the following elements:
 
 ``Mutation`` type with two mutations: ``login`` mutation that requires username and password strings and returns bool with status, and ``logout`` that takes no arguments and just returns status.
 
-For the sake of simplicity, our mutations return bools, but really there is no such restriction. You can have a resolver that returns status code, an updated object, or an error message::
+For the sake of simplicity, our mutations return bools, but really there is no such restriction. In fact, it is recommended for mutations to return dedicated ``Payload`` types:
```
rafalp: Missing double colon (`::`) at the end of paragraph.

Also, lets replace "there is no such restriction" with "there is no such requirement", and second part to `it is recommended as good practice that mutations return dedicated "payload" types::`

docs/mutations.rst
```diff
@@ -97,11 +100,13 @@ Imagine a mutation for creating ``Discussion`` that takes category, poster, titl
...
         type Mutation {
-            createDiscussion(category: ID!, title: String!, isAnnouncement: Boolean, isClosed: Boolean) {
-                status: Boolean!
-                error: Error
-                discussion: Discussion
-            }
+            createDiscussion(category: ID!, title: String!, isAnnouncement: Boolean, isClosed: Boolean): DiscussionPayload
+        }
+
+        type DiscussionPayload {
```
rafalp: IMHO, we shouldn't add payload for this (and next)  example - we've already explained payload concept in the docs, and those examples focused on illustrating `input` usage.

docs/mutations.rst
```diff
@@ -97,11 +100,13 @@ Imagine a mutation for creating ``Discussion`` that takes category, poster, titl
 
     type_def = """
         type Mutation {
-            createDiscussion(category: ID!, title: String!, isAnnouncement: Boolean, isClosed: Boolean) {
-                status: Boolean!
-                error: Error
-                discussion: Discussion
-            }
+            createDiscussion(category: ID!, title: String!, isAnnouncement: Boolean, isClosed: Boolean): DiscussionPayload
```
rafalp: We could reformat this part to remove horizontal scroll:

```
createDiscussion(
    category: ID!,
    title: String!,
    isAnnouncement: Boolean,
    isClosed: Boolean
): DiscussionPayload
```

docs/mutations.rst
```diff
@@ -33,18 +33,21 @@ In this example we have the following elements:
 
 ``Mutation`` type with two mutations: ``login`` mutation that requires username and password strings and returns bool with status, and ``logout`` that takes no arguments and just returns status.
 
-For the sake of simplicity, our mutations return bools, but really there is no such restriction. You can have a resolver that returns status code, an updated object, or an error message::
+For the sake of simplicity, our mutations return bools, but really there is no such restriction. In fact, it is recommended for mutations to return dedicated ``Payload`` types::
```
rafalp: Let's reword the "there is no such restriction" to "there is no such requirement", and second part to `it is generally considered a good practice for mutations to return dedicated "payload" types`

Also we could turn this bit into sub-section so its easy to refer to later. Maybe "Return values" or "Mutation payloads"?

PR #60, opened 2018-11-15
Add support for query batching

patrys:
⚠️ Work in progress!

TODO:
* [x] Turn WSGI middleware into a generator
* [x] Support query batching
* [ ] Add tests

### DIFF
--- a/ariadne/wsgi_middleware.py
+++ b/ariadne/wsgi_middleware.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Callable, List, Optional, Union
+from typing import Any, Callable, Generator, List, Optional, Tuple, Union
 from wsgiref import simple_server
 
 from graphql import format_error, graphql
@@ -21,6 +21,11 @@ from .exceptions import (
     HttpMethodNotAllowedError,
 )
 from .executable_schema import make_executable_schema
+
+Query = dict
+BatchQuery = Union[Query, List[Query]]
+BatchExecutionResult = Union[ExecutionResult, List[ExecutionResult]]
+WSGIResponse = Generator[bytes, None, None]
 
 
 class GraphQLMiddleware:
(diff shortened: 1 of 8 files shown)


jacobmischka:
Is there any update on this? I searched through the docs and it doesn't look like this was ever implemented, is that right? Without query batching in some way it seems like ariadne suffers from the dreaded n+1 query problem, is that right? Are there any workarounds I missed?

Thanks!

rafalp:
Hi @jacobmischka, thanks for interest!

This feature is currently on hold, but we we will eventually return to it and get it implemented.

> Without query batching in some way it seems like ariadne suffers from the dreaded n+1 query problem, is that right?

Query batching is not a solution to the N+1 problem. It's solution to the "I want to run few queries/operations in single trip to the server" problem that especially mobile clients have.

Ariadne already supports two most popular solutions to the N+1 problem:

- you can provide [root value resolver](https://ariadnegraphql.org/docs/types-reference#rootvalue) that introspects GraphQL query and loads it in optimized way
- if your python code is asynchronous, you can implement custom dataloader object that returns [futures](https://docs.python.org/3/library/asyncio-future.html) to your resolvers and after some delay (say, 10ms) loads the data to those futures. This strategy is most popular in node.js, but its also viable if you are running on asynchronous stack.

jacobmischka:
Ohhh that's right, I forgot about that graphql feature. Perfect, thanks much!

patrys:
You don't need to write custom data-loaders, there's https://github.com/syrusakbary/aiodataloader

dacevedo12:
any update on this? It looks really cool. Is It going to be supported for ASGI as well?

rafalp:
@dacevedo12 remember that QueryBatching is a mechanism used to save time sending multiple queries in single HTTP request. It doesn't affect query executor so it makes no difference if you are using ASGI or WSGI for your server.

patrys:
With HTTP/2+ it's not as important as it was with HTTP/1.x so I'm no longer sure whether it's worth finishing this pull request given how complex it makes the query handling logic.

ShadyNawara:
> https://github.com/syrusakbary/aiodataloader

Is there any usage samples for this with ariadne?

ShadyNawara:
> > https://github.com/syrusakbary/aiodataloader
> 
> Is there any usage samples for this with ariadne?

found that you generously provided an example here https://spectrum.chat/ariadne/general/make-single-call-to-service-that-provides-data-to-multiple-resolvers~d8ba151d-ceab-4771-b8c5-bbf628fb3eb9?m=MTU2Nzc4NzI3Mjg1Mg==

nikordaris:
@rafalp 

> Query batching is not a solution to the N+1 problem. It's solution to the "I want to run few queries/operations in single trip to the server" problem that especially mobile clients have.

Isn't query batching required for Apollo Federation batching of keys? so if you are trying to protect against N+1 at the Gateway level, then you need Query Batching + DataLoader to mitigate the N+1 problem created by a Federated system?

rafalp:
@process0 I think we could have feature flag to enable query batching, but first I want to get #855 merged

process0:
@rafalp now that #855 is merged, what is required for this?

rafalp:
@process0 question about federation doesn't seem accurate. Federated servers implement `_entities` query that receives list of type:id pairs of items it should return to federation gateway, so there's no N+1 problem occurring here.

As for batching executor, I would like to see a PR that implements this as batching HTTP handler replacing default one.

feslima:
@rafalp I'm currently trying to implement this approach of using the list of type:id pairs that comes from the federation gateway. Do you have any resources/examples on how can I get this values from inside the resolvers? I'm using flask + ariadne and going async is not an option :(

I've tried [this using a dataloader with a WSGI](https://github.com/mirumee/ariadne/discussions/915#discussioncomment-3438068), but it didn't work. When the resolver loads the entity from the instantiated data loader, it returns a Future instance instead of the instance in memory.

rafalp:
Hello Felipe,
Good news is next week I would like to release the Ariadne 0.18 which will include the dataloaders support for synchronous servers. :)

feslima:
@rafalp Thank you very much for the response and good news! I'll be waiting for it.

anabeatrizgn:
@rafalp excited for this new release too! waiting for this dataloaders support for synchronous servers

rafaelalexandrino:
@rafalp Waiting for this new release! That is a good new!

leomontenegro2104:
That's great news, @rafalp. Looking forward to the release of this new version

rafalp:
I don't know why this discussion is happening in query batching PR 😅

Anyway, I'm sorry but there will be delays with next release, biggest blocker is finishing docs for 0.18, but [dataloaders documentation is already there](https://ariadnegraphql.org/docs/next/dataloaders), so I guess we'll have 0.18.0 beta release before proper final release?

rafalp:
I've marked dataloaders dicussion as offtopic.

PR #61, opened 2018-11-15
Migrate to graphql-core-next

rafalp:
Now that Ariadne requires python 3.6 to run, we can migrate our codebase to graphql-core-next, bringing us some cool new things we could use:

- `build_schema` that replaces Ariadne's implementation for building schema from SDL
- async support for stacks that can use it
- support for more current version of GraphQL specification (directives and descriptions yay!)

### TODO

- [x] Make tests pass
- [x] Update custom scalars docs (raise Value/Type exception instead of returning `None`)
- [x] Update `ResolveInfo` documentation
- [x] Update changelog

### TESTS

- [x] `test_queries`
- [x] `test_mutations`
- [x] `test_enums`
- [x] `test_custom_scalars`
- [x] `test_modularization`
- [x] `test_introspection`

#### WSGI Middleware

- [x] `test_custom_middleware`
- [x] `test_http_error_handling`
- [x] `test_initialization`
- [x] `test_playground`
- [x] `test_query_execution`
- [x] `test_request_data_reading`
- [x] `test_request_dispatching`
- [x] `test_simple_server`


CHANGELOG.md
```diff
@@ -2,4 +2,8 @@
 
 ## 0.2.0 (UNRELEASED)
 
-- Removed support for Python 3.5 and added support for 3.7
\ No newline at end of file
+- Removed support for Python 3.5 and added support for 3.7.
+- Moved to `GraphQL-core-next` that supports `async` resolvers, query execution and implements more recent version of GraphQL spec. If you are updating existing project, you will need to uninstall `graphql-core` before installing `graphql-core-next`, as both libraries use `graphql` namespace.
+- `Boolean` build-in scalar now checks the type of serialized value. Returning value of type other than `bool`, `int` or `float` from field resolver will result in `Boolean cannot represent a non boolean value` error.
```
patrys:
```suggestion
- `Boolean` built-in scalar now checks the type of each serialized value. Returning values of type other than `bool`, `int` or `float` from a field resolver will result in a `Boolean cannot represent a non boolean value` error.
```

ariadne/wsgi_middleware.py
```diff
@@ -169,9 +182,8 @@ def return_response_from_result(
         status = HTTP_STATUS_200_OK
         response = {}
         if result.errors:
-            response["errors"] = [format_error(e) for e in result.errors]
-        if result.invalid:
             status = HTTP_STATUS_400_BAD_REQUEST
```
patrys: Do we want to return 400 each time an error occurs? `apollo-server` will only return 400 if an error is thrown while parsing the query and/or variables. In such cases the response contains just the `errors` field and not `data`. In all other cases the status code is 200 and the `data` field is present (even if set to `null`).

tests/test_custom_scalars.py
```diff
@@ -24,23 +24,23 @@ def serialize(date):
...
+    if not isinstance(ast, StringValueNode):
+        raise ValueError()
 
     formatted_date = ast.value
     try:
         parsed_datetime = datetime.strptime(formatted_date, "%Y-%m-%d")
         return parsed_datetime.date()
     except ValueError:
-        pass
+        raise ValueError()
```
patrys: Why catch and raise the same exception?

rafalp: I've updated the docs to provide explanation:

> If we didn't swap the original exception, it's message would be included by the GraphQL at the end of generated message, possibly **leaking implementation details** to the client.

ariadne/wsgi_middleware.py
rafalp: No opinion here. Originally we had nice `invalid` flag on the result we could've used to make the call. Now that it's gone, we could return 400 for errors coming from `validate_query`, and always return 200 when response is coming from `return_response_from_result`.

There's a gotcha here is that we can't reliably check `operationName` without looking into the parsed query's AST, so passing operation name thats string, but doesn't exist in the query could produce 200 with `"data": "null"`, but `errors` containing the message.

rafalp: I've changed this so we return 400 only for `validate_query` errors.

tests/test_custom_scalars.py
rafalp: I've updated our tests and docs.

NyanKiyoshi: Couldn't we add a Python comment as well?

tests/wsgi_middleware/test_query_execution.py
```diff
@@ -1,4 +1,5 @@
 # pylint: disable=too-many-arguments
+from ariadne.constants import HTTP_STATUS_200_OK, HTTP_STATUS_400_BAD_REQUEST
```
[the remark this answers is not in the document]
rafalp: I've turned this into new issue.

We also want to extract parts of `GraphQLMiddleware` into reusable utility functions so we can share some view logic between Ariadne and other libs (eg. Django views). We may move those constants there by then.

Thing to keep on mind is that currently only constants Ariadne has are HTTP statuses and content types.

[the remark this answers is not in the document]
salwator: Maybe we just don't need generic _constants_ namespace 😄

PR #74, opened 2018-11-27
Move library API to objects

rafalp:
This PR replaces current procedural API inspired by Apollo-server with more *pythonic* object API.

This reverses order in which resolvers are defined and mapped to type, saving us gotchas from forgetting about updating resolvers map:

```python
schema = Schema(type_defs)

query = schema.type("Query")

@query.resolve("persons")
def resolve_persons(*_):
    return []
```

As usual, this is a work in progress, and there will be dragons ;)

### TODO

- [ ] Finalize API
- [ ] Write tests
- [ ] Update docs
- [ ] Update readme
- [ ] Nuke old unused code
- [ ] Update changelog

### DIFF
--- a/ariadne/__init__.py
+++ b/ariadne/__init__.py
@@ -1,14 +1,18 @@
 from .executable_schema import make_executable_schema
 from .resolvers import add_resolve_functions_to_schema, default_resolver, resolve_to
-from .utils import gql, start_simple_server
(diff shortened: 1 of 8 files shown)


rafalp:
We've discussed this approach internally some more, and settled that the previous one is more promising.

PR #75, opened 2018-12-11
Enable schema loading from file or directory

### LINKED ISSUE #62: Provide utility for loading types from file/directory
We could provide a utility function that would take a path to the filesystem, and:

- if its a file, read it, validate that its correct file, and return SDL string
- if its a directory, glob `.graphql` files, validate each file, and concat them into single SDL string

That way users could easily use single schema definition for both graphql API and frontend tools.

salwator:
Closes #62

From now on, you can use _load_schema_from_path()_ function to load GraphQL schema from file or directory of _.graphql_ files. Just provide provide path, schema will be parsed and joined, syntax errors are presented with information about source file.

### TODO

* [x] Load single file
* [x] Glob files from directory
* [x] Update `CHANGELOG.md`


ariadne/__init__.py
```diff
@@ -24,6 +26,7 @@
     "fallback_resolvers",
     "gql",
     "make_executable_schema",
+    "load_schema_from_path",
```
rafalp: `load` should be defined before `make`

salwator: Sure

PR #90, opened 2019-01-09
Support transforming enum values to internal representation

### LINKED ISSUE #39: Support custom internal representation of enums
Currently Ariadne supports enums that are represented as strings that are mapped 1:1 to what they are in SDL. This should take care of majority of use cases, but there may be edge cases and legacy systems, where enum is defined for existing value that may be integer, and updating python implementation is out of question.

[Apollo doc for reference.](https://www.apollographql.com/docs/apollo-server/features/scalars-enums.html#internal-values)

rafalp:
This PR adds `Enum` *bindable* that takes name of enum in schema and dict of mappings, enabling easy transformation of enum values between GraphQL and internal representation:

```
role = Enum(
    "role",
    {
        "ADMIN": 1,
        "MODERATOR": 2,
        "MVP": 4,
        "REGULAR" 5,
        "PROBATION": 6,
    }
)
```

Fixes #39

### DIFF
--- a/tests/test_enums.py
+++ b/tests/test_enums.py
@@ -87,7 +87,7 @@ def test_attempt_bind_custom_enum_to_schema_enum_missing_value_raises_error(
 ):
     enum = Enum("Episode", {"JARJAR": 1999})
     with pytest.raises(ValueError):
-        enum.bind_to_schema(schema_with_enum)
+        enum.bind_to_schema(schema_with_enum)  # pylint: disable=no-member
 
 
 custom_enum = Enum("Episode", {"NEWHOPE": 1977, "EMPIRE": 1980, "JEDI": 1983})
--- a/tests/test_enums.py
+++ b/tests/test_enums.py
@@ -1,3 +1,5 @@
+import enum
+
 import pytest
 
 from graphql import graphql_sync, build_schema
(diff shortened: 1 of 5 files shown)


rafalp:
Note to self: passing actuall `Enum` as `ariadne.Enum` value should be possible too:

```
my_enum = MyEnum(Enum):
    RED = 1
    BLUE = 2


graphql_enum = ariadne.Enum("SomeEnum", my_enum)
```

rafalp:
And now you can also pass either `enum.Enum` or `enum.IntEnum` to `ariadne.Enum`.

tests/test_enums.py
```diff
@@ -45,7 +49,7 @@ def test_unsuccesfull_invalid_enum_value_evaluation():
 """
 
 
-def test_succesfull_enum_value_passed_as_argument():
+def test_successfull_enum_value_passed_as_argument():
```
patrys:
```suggestion
def test_successful_enum_value_passed_as_argument():
```

tests/test_enums.py
```diff
@@ -54,10 +58,112 @@ def test_succesfull_enum_value_passed_as_argument():
     assert result.errors is None, result.errors
 
 
-def test_unsuccesfull_invalid_enum_value_passed_as_argument():
+def test_unsuccessfull_invalid_enum_value_passed_as_argument():
```
patrys:
```suggestion
def test_unsuccessful_invalid_enum_value_passed_as_argument():
```

PR #94, opened 2019-01-20
Fix typo in README

commandtab:
`type_defs` defines a `fullName` field, but the resolver field is named `fullname` (no capital N), raising a ValueError. This PR fixes the typo and makes the example run.


rafalp:
Nice catch, thank you! 👍

PR #96, opened 2019-01-23
Update Django integration docs

bogdal:
This PR makes the example code runnable with the default Django configuration.


rafalp:
Not sure if we'll won't remove this doc down the line (we plan to have proper `ariadne.contrib.django` in the lib) but merging in meantime this improvement won't hurt.

Thanks 👍

PR #97, opened 2019-01-25
Add subscriptions

patrys:
As simple as that. Subscriptions.

Bear in mind that as long as the example server is synchronous, the subscriptions won't work with the example server.

I'm not a fan of the API we're ending up with for the `ResolverMap`.

I think I'd prefer:

```python
sub_map = ResolverMap("Subscription")

@sub_map.add_field("counter")
def counter(count, info):
    return count

@sub_map.add_source("counter")
async def tick(root, info):
    ...
```

They also look better when used as non-decorators:

```python
sub_map.add_field("counter", resolver=counter)
sub_map.add_source("counter", generator=tick)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.3.0 (UNRELEASED)
 
 - Added `Enum` type for mapping enum variables to internal representation used in application.
+- Added support for subscriptions.
 
 ## 0.2.0 (2019-01-07)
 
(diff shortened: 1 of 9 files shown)


patrys:
The source (I really dislike the term "subscriber", the client is the subscriber) yields things. In real scenarios they will more often than not have a shape that does not match that of the resolver. I imagine most sources will yield things like ORM records that may not be a 1:1 match of what you're exposing to the world.

I like the idea of having `@field(...)` and `@source(...)` for decorators and `add_field`/`add_source` for regular use. This matches what Starlette is doing.

rafalp:
We also have the `Following features should work but are not tested and documented: unions, interfaces and subscriptions.` bit in readme and on docs index that will need updating.

patrys:
Should I add coverage pragma to the generators in tests?

rafalp:
Yup, lets add those.

PR #102, opened 2019-02-02
[docs/modularization] Fix missing close quotation mark

ajhyndman:
Just a quick typo fix I spotted while reading docs.

(Also, I'm really excited to have discovered `ariadne`, it looks great!)


rafalp:
Great catch, thanks!

Issue #98, opened 2019-01-26
Change procedural API for adding functions to fields to `add_field`/`add_source`


rafalp:
Currently Ariadne `ResolverMap` `field` attribute can be used both as decorator and function:

```
@resolver_map.field("fieldname")
def some_resolver():
    ...


resolver_map.field("fieldname", resolver=some_resolver)
```

This name works when used as decorator, but is confusing when procedural approach is used by dev instead. Because we have two uses for same function, we also have complex implementation and type annotations. Splitting `field` into two functions would greatly simplify things for us.

We would also have to check `Scalar` and update it accordingly (but perhaps use `set_` prefix?)

CC @patrys

rafalp:
For API consistency we should also update the `Enum` implementation to let you create enum via `set_items` perhaps.

patrys:
Shouldn't the enum also accept all of its attributes in the constructor? Compare:

```python

foo = Enum("Foo")
foo.set_items(something)

# vs.

foo = Enum("Foo", something)
```

patrys:
Slightly related: should we warn if you try to set items/resolver/source/whatever of the same type for a second time using the same bindable instance?

rafalp:
> Shouldn't the enum also accept all of its attributes in the constructor? Compar

This is correct, and its already the case with our Enums.

Likely we should update `Scalar` to accept three `kwargs` in addition to decorators.

The question is if there are scenarios out there for using procedural API for setting scalar resolvers, ergo:

```python
foo = Scalar("test")
foo.set_serializer(my_serializer)
```

This looks good on the surface, but what possible uses are there for it?

> Slightly related: should we warn if you try to set items/resolver/source/whatever of the same type for a second time using the same bindable instance?

Yup, missing the resolver on field can lead to tricky bugs. We should signalize to the dev if such situations happen.

rafalp:
I've decided to sit down and put together the draft for API change:

### ResolverMap

```python
foo = ResolverMap("Name")


# decorator syntax
@foo.field("name")
def resolve_name(*_):
    ...


# setter syntax
foo.set_alias("name", "attr_name")
foo.set_field("name", resolve_name)
foo.set_source("name", observe_name)
```


### Scalar

Decorator syntax:

```python
foo = Scalar("Name")


@foo.serializer
def serialize_foo(*_):
    ...


@foo.value_parser
def parse_foo_value(*_):
    ...


@foo.literal_parser
def parse_foo_literal(*_):
    ...
```

[...]

PR #107, opened 2019-02-05
Add unions

rafalp:
This PR adds support for `union` types in GraphQL.

This is achieved by introduction of `ariadne.Union` bindable that takes name of union, and resolver function that takes any object and is expected to return `None` or `str` with GraphQL type name.

`Union` also implements new API as discussed in #98:

```
union = Union("Name", type_resolver)

# ...or...

@union.type_resolver
def resolver(*_):
    ...
```

### DIFF
--- a/tests/test_unions.py
+++ b/tests/test_unions.py
@@ -59,9 +59,9 @@ thread = Mock(title="Thread")
 
 def test_union_type_resolver_may_be_set_on_initialization():
     query = ResolverMap("Query")
-    query.field(
+    query.field(  # pylint: disable=unexpected-keyword-arg
         "item", resolver=lambda *_: user
-    )  # pylint: disable=unexpected-keyword-arg
+    )
 
     union = Union("FeedItem", type_resolver=lambda *_: "User")
     schema = make_executable_schema(type_defs, [query, union])
@@ -72,9 +72,9 @@ def test_union_type_resolver_may_be_set_on_initialization():
 
 def test_union_type_resolver_may_be_set_using_decorator():
     query = ResolverMap("Query")
-    query.field(
+    query.field(  # pylint: disable=unexpected-keyword-arg
         "item", resolver=lambda *_: user
-    )  # pylint: disable=unexpected-keyword-arg
+    )
 
     union = Union("FeedItem")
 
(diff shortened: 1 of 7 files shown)


tests/test_unions.py
```diff
@@ -0,0 +1,132 @@
...
+                username
+            }
+            ... on Thread {
+                title
+            }
+        }
+    }
+"""
+
+user = Mock(username="User")
```
salwator: Naming module-scoped symbols should be different then local variables, especially when we have same naming convention for pytest fixtures.
I would make it _USER_ or _User_ at least. Same applies for thread.

tests/test_unions.py
```diff
@@ -0,0 +1,132 @@
...
+    result = graphql_sync(schema, "{ item { __typename } }")
+    assert result.data == {"item": {"__typename": "User"}}
+
+
+def resolve_result_type(obj, *_):
+    if obj == user:
+        return "User"
+    if obj == thread:
+        return "Thread"
+    return None
```
salwator: No need for that explicit return.

tests/test_unions.py
```diff
@@ -0,0 +1,132 @@
...
+    union = Union("FeedItem", type_resolver=resolve_result_type)
+
+    schema = make_executable_schema(type_defs, [query, union])
+    result = graphql_sync(schema, test_query)
+    assert result.data == {"item": {"__typename": "Thread", "title": thread.title}}
+
+
+def test_result_is_none_if_union_didnt_resolve_the_type():
+    query = ResolverMap("Query")
+    query.field(  # pylint: disable=unexpected-keyword-arg
```
salwator: We might want to just disable warning for tests, for example by local pylintrc file.

salwator: Or we can just `# pragma pylint: disable=unexpected-keyword-arg` at the top.

tests/test_unions.py
```diff
@@ -0,0 +1,132 @@
...
+    if obj == user:
+        return "User"
+    if obj == thread:
+        return "Thread"
+    return None
+
+
+def test_result_is_username_if_union_resolves_type_to_user():
+    query = ResolverMap("Query")
+    query.field(  # pylint: disable=unexpected-keyword-arg
```
salwator: Resolver map with configured field is duplicated in test cases, it is perfect case for pytest fixture.

tests/test_unions.py
rafalp: I agree that code littered with comments disabling linter looks bad, but I think it's more explicit and safer to work with than one that blanket-disables everything.

I'm also using those as a reminder that our current `ResolverMap` API confuses the hell out of `PyLint` and we really need to change it, but that's tracked in #98

tests/test_unions.py
rafalp: This code is here to satisfy pylint `inconsistent-return-statements`.

PR #105, opened 2019-02-04
Document Ariadne's support for Descriptions [initial draft]

ajhyndman:
Hey guys!  I tried my hand at a first draft documenting Ariadne's support for the GraphQL descriptions feature.

I didn't go into much detail, but declared that the feature is supported, and gave an example of basic usage.  Does it feel like I'm on the right track?

Fixes #54

Issue #54, opened 2018-10-31
API descriptions


rafalp:
GraphQL supports item descriptions, but currently, Ariadne provides no way to set those, and neither does `GraphQL-Core` version we are using.

Ideally, we should provide two ways to set item descriptions:

- if resolver has docstring, we should use it
- add `description=` kwarg to `make_executable_schema` & friends that would take dict of dicts and would override items descriptions based on that. We could read special key (eg. `__description`) to get description for type. This approach should also support modularization.

patrys:
I'm all for making descriptions part of schema, not implementation. Using mock resolvers should not alter the results of introspection.

rafalp:
Ok. As descriptions are already implemented by GraphQL Core Next, closing this issue should involve adding documentation for descriptions.

PR #109, opened 2019-02-06
Test the GraphQL descriptions feature

ajhyndman:
This is required for #54.

This is also a prerequisite for #105.

@rafalp Again, let me know if I'm on the right track, or if you'd like me to expand on anything!

### DIFF
--- a/tests/test_descriptions.py
+++ b/tests/test_descriptions.py
@@ -57,7 +57,7 @@ def test_enum_has_description(schema):
 
 
 def test_enum_value_has_description(schema):
-    value = schema.get_type("Language").values["EN"]
+    value = schema.get_type("Language").values.get("EN")
     assert isinstance(value.description, str)
 
 
@@ -73,7 +73,7 @@ def test_object_field_has_description(schema):
 
 def test_object_field_argument_has_description(schema):
     translate_field = schema.get_type("Query").fields.get("translate")
-    argument = translate_field.args["fromLanguage"]
+    argument = translate_field.args.get("fromLanguage")
     assert isinstance(argument.description, str)
 
 
--- a/tests/test_descriptions.py
+++ b/tests/test_descriptions.py
@@ -35,6 +35,14 @@ type_defs = '''
         "Chinese"
         CH
     }
+
+    """
+    A poorly documented object type.
+    """
+    type MyObject {
+        myField: String
+        transform(a: String): String
+    }
 '''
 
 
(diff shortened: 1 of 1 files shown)


tests/test_descriptions.py
```diff
@@ -0,0 +1,78 @@
...
+    description = schema.get_type("Query").description
+    assert description == "A simple GraphQL schema which is well described."
+
+
+def test_object_fields_include_descriptions():
+    schema = make_executable_schema([query_definition, enum_definition])
+
+    description = schema.get_type("Query").fields.get("translate").description
+    assert (
+        description
```
rafalp: Is this formatted by black? Could you replace the line with `description == ("multiline string")` and see if its better?

tests/test_descriptions.py
```diff
@@ -0,0 +1,78 @@
...
+
+    description = schema.get_type("Language").description
+    assert description == "The set of languages supported by `translate`."
+
+
+def test_enum_values_include_descriptions():
+    schema = make_executable_schema(enum_definition)
+
+    for value in schema.get_type("Language").values.values():
+        assert isinstance(value.description, str)
```
rafalp: Could this be replaced with 1 assert explicitly testing one of items for having description?

Loops generate nice coverage, but they provide any feedback if they didn't loop over anything, which is bad. :(

tests/test_descriptions.py
```diff
@@ -0,0 +1,78 @@
...
+        description
+        == "Translates a string from a given language into a different language."
+    )
+
+
+def test_object_field_arguments_include_descriptions():
+    schema = make_executable_schema([query_definition, enum_definition])
+
+    translate_field = schema.get_type("Query").fields.get("translate")
+
```
rafalp: I would drop this blank line. Doesn't seem like it adds much to readability here imho.

tests/test_descriptions.py
```diff
@@ -0,0 +1,78 @@
...
+    for argument in translate_field.args.values():
+        assert isinstance(argument.description, str)
```
rafalp: Same feedback as for one above. Please replace loop with explicit assert.

ajhyndman:
Thanks for the detailed feedback!

Sorry for the slow turnaround, I promise I don't mean to drag my feet.  Updating this just hasn't made it to the top of my todo list every day.

rafalp:
Great work! Thanks!

PR #112, opened 2019-02-19
Add interfaces

rafalp:
This PR adds `Interface` type to Ariadne. Interface acts like the union, except it enforces intersection between implementing types. It also allows (optional) resolvers reuse between implementing types.

~~Currently, this is a proof-of-concept implementation.~~

### TODO

* [x] Use interface to reuse resolvers
* [x] Use interface as return type (needs `type_resolver` like union does)
* [x] Tests
* [x] Docs
* [x] Changelog

### DIFF
--- a/docs/interfaces.rst
+++ b/docs/interfaces.rst
@@ -40,7 +40,7 @@ Type definitions can then be updated to ``implement`` this interface::
     }
 
 
-GraphQL standard requires that every type implementing the ``Interface`` also explicitly defines the fields that this interface defines. This is why the ``summary`` and ``url`` fields repeat on all types in the example.
+GraphQL standard requires that every type implementing the ``Interface`` also explicitly defines fields from the interface. This is why the ``summary`` and ``url`` fields repeat on all types in the example.
 
 Like with the union, the ``SearchResult`` interface will also need a special resolver named *type resolver*. This resolver will we called with an object returned from a field resolver and current context, and should return a string containing the name of a GraphQL type, or ``None`` if the received type is incorrect::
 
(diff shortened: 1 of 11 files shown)


docs/interfaces.rst
```diff
@@ -0,0 +1,105 @@
+Interface types
+===============
+
+Just like previously described :ref:`unions <unions>`, interfaces enable you to create fields that resolve to a value that is one of a few possible types.
```
salwator: Should we really describe interfaces in comparison to unions? I know that from library implementation point of view they share some implementation details, but idea is wildly different.
I would start with defining interface as an abstract type, similar to those available in other languages, that enforces on types implementing it existence of defined fields and enable polymorphic behaviors in API. For example, one might return list of objects that shares common features.

docs/interfaces.rst
```diff
@@ -0,0 +1,105 @@
...
+-------------------------------
+
+Ariadne's ``Interface`` instances can also optionally be used to share resolvers between implementing types::
+
+    @search_result.field("summary")
+    def resolve_summary(obj, *_):
+        return str(obj)
+
+    
+    @search_result.field("url")
```
salwator: Can we have an actual example of _sharing_ here?

tests/test_interfaces.py
```diff
@@ -0,0 +1,246 @@
...
+    }
+"""
+
+
+@pytest.fixture
+def schema():
+    return build_schema(type_defs)
+
+
+def test_attempt_bind_interface_to_undefined_type_raises_error(schema):
```
salwator: attempt *to*

docs/interfaces.rst
rafalp: I agree. Care to create a change suggestion?

rafalp:
@salwator I've made another pass over the docs. Could you give it a look and let me know what you think about it now?

PR #117, opened 2019-03-19
Library API cleanup

### LINKED ISSUE #98: Change procedural API for adding functions to fields to `add_field`/`add_source`

rafalp:
This PR passes over library API fixing headscratchers like `field` working both as decorator and method taking two arguments which confused code analysis tools, as well as attempting to make our utility classes names more predictable, searchable and don't clash with already existing names (eg. `ariadne.Enum` vs. `enum.Enum`)

Fixes #98 

### Objects naming

- [x] Rename `ResolverMap` to `ObjectType`
- [x] Split subscriptions from `ResolverMap` to `SubscriptionType`
- [x] Create `QueryType`
- [x] Create `MutationType`
- [x] Rename `Enum` to `EnumType`
- [x] Rename `Interface` to `InterfaceType`
- [x] Rename `Scalar` to `ScalarType`
- [x] Rename `Union` to `UnionType`
- [x] Rename `Bindable` base class to `SchemaBindable`
- [x] Update quick start example in readme
- [x] Update existing documentation
- [x] Unify docs and arguments names for ``SchemaBindable``s
- [x] Document `SchemaBindable`

### Object API

- [x] Split `field` into `field` decorator and `set_field` setter
- [x] Rename `alias` to `set_alias`
- [x] Split `source` into `source` decorator and `set_source` setter
- [x] Add `serializer`, `value_parser` and `literal_parser` kwargs to `Scalar` constructor


patrys:
One nitpick: I'd consider renaming the Bindable to SchemaBindable rather than BindableType as its mainly a mechanism for calling bind_to_schema.

rafalp:
@patrys ideally I would like to have a nice name that we could also use for arg name for `make_executable_schema` and friends wrapping it, currently we have `resolvers` which is awkward because we have plenty other objects we are telling users to pass to it that ain't resolvers.

`SchemaBindable` -> `bindables` looks good. We could also then name concept "Schema bindables" and have doc for implementing custom "bindables" explaining how to implement custom abstractions in place of default ones.

rafalp:
I've couldn't get it to be 100% consistent on `resolvers` vs `bindables`, but I feel what we've got for now is okay.

PR #119, opened 2019-03-25
Add ASGI module to Ariadne

patrys:
Work in progress implementation of an ASGI server based on Starlette.


README.md
```diff
@@ -82,9 +93,14 @@ def resolve_person_fullname(person, *_):
 
 # Create executable GraphQL schema
 schema = make_executable_schema(type_defs, [query, person])
+# Create an ASGI app
```
rafalp:
```suggestion

# Create an ASGI app using the schema
```

ariadne/asgi.py
```diff
@@ -0,0 +1,230 @@
...
+GQL_STOP = "stop"  # Client -> Server
+
+
+class GraphQL:
+    def __init__(self, schema: GraphQLSchema, *, keepalive: float = None):
+        self.keepalive = keepalive
+        self.schema = schema
+
+    def __call__(self, scope: Scope):
+        assert scope["type"] in {"http", "websocket"}
```
rafalp: We should remove either assert or `ValueError` from end of this method

ariadne/asgi.py
```diff
@@ -0,0 +1,230 @@
...
+                payload["errors"] = [format_error(e) for e in result.errors]
+            await self.send_json(
+                websocket, {"type": GQL_DATA, "id": operation_id, "payload": payload}
+            )
+        await self.send_json(websocket, {"type": GQL_COMPLETE, "id": operation_id})
+
+    async def keep_connection_alive(self, websocket: WebSocket):
+        if not self.keepalive:
+            return
+        while True:
```
rafalp:
```suggestion
        while websocket.application_state != WebSocketState.DISCONNECTED::
```

We can drop bottom `if/break` that way

ariadne/asgi.py
```diff
@@ -0,0 +1,230 @@
...
+
+    async def send_json(self, websocket, message) -> None:
+        message = json.dumps(message)
+        await websocket.send_text(message)
+
+    async def graphql_ws_server(self, websocket: WebSocket) -> None:
+        subscriptions: Dict[str, AsyncGenerator] = {}
+        await websocket.accept("graphql-ws")
+        try:
+            while True:
```
rafalp: Could we break this block down into smaller methods?

docs/modularization.rst
```diff
@@ -28,8 +29,9 @@ To load schema from file or directory, you can use the ``load_schema_from_path``
     # build an executable schema
     schema = make_executable_schema(type_defs)
 
-    # Start server that can't execute any queries, but allows you to browse your schema
-    start_simple_server(schema)
+    app = GraphQL(schema)
```
rafalp: Move this below next two comments and reword those to something like "Use schema to construct GraphQL app for an ASGI server"?

patrys:
Checks currently fail because `black` is configured to check generated files, @rafalp will take care of that.

PR #122, opened 2019-03-27
Use poetry for package

rafalp:
With PIP 19 being released, we can safely move Ariadne to Poetry removing the need for maintaining separate `requirements*` files or `setup.py`.

### TODO

* [x] Create pyproject.toml
* [ ] Add requirements using Poetry
* [ ] Remove `setup.py`
* [ ] Remove `requirements` files
* [ ] Update Travis config
* [ ] Add config for black to exclude `snapshot` directory

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,34 @@
+[build-system]
+requires = ["poetry"]
+build-backend = "poetry.masonry.api"
+
+[tool.poetry]
+name = "ariadne"
+version = "0.3.0"
+description = "Ariadne is a Python library for implementing GraphQL servers."
+authors = ["Mirumee Software <hello@mirumee.com>"]
+
+[tool.poetry.dependencies]
+python = ">=3.6"
+graphql-core-next = "^1.0"
+typing = "^3.6"
+typing_extensions = "^3.7"
+
+
+[tool.poetry.dev-dependencies]
+black = { version = "*", allows-prereleases = true }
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+description = "Ariadne is a Python library for implementing GraphQL servers."
+authors = ["Mirumee Software <hello@mirumee.com>"]
+
+[tool.poetry.dependencies]
+python = ">=3.6"
+graphql-core-next = "^1.0"
+typing = "^3.6"
+typing_extensions = "^3.7"
+
+
```
NyanKiyoshi: Did a oopsie here:
```suggestion
```

rafalp:
@NyanKiyoshi from what I know `pyproject.toml` replaces `setup.py` altogether, with PIP 19 reading the build tool from it and following from there with poetry.

nikordaris:
FWIW, I've recently abandoned Poetry to go back to requirements files because of frequent breaking changes between pip and poetry. The unexpected thing was these issues would appear at package install time in my apps. This forced my apps to downgrade pip to compatible versions until the lib got upgraded to a newer version of poetry that was compatible with the latest pip which was unacceptable for a tool I viewed as a dev only packaging convenience.

patrys:
We're not planning to drop `requirements.txt`, just change the tool we use to maintain it.

Issue #125, opened 2019-03-31
Custom Server example with Flask


jrast:
Hey there! I ported the custom server example from django to flask.

It's quite short and it would be nice to add it to the docs.

````python
from ariadne import gql, ResolverMap, make_executable_schema
from ariadne.constants import PLAYGROUND_HTML
from graphql import format_error, graphql_sync
from flask import request, jsonify
from flask.views import MethodView

type_defs = gql("""
    type Query {
        hello: String!
    }
""")


query = ResolverMap("Query")


@query.field("hello")
def resolve_hello(_, info):
    request = info.context
    print(request.headers)
    user_agent = request.headers.get("User-Agent", "Guest")
    return "Hello, %s!" % user_agent


schema = make_executable_schema(type_defs, query)


class GraphQlView(MethodView):

    def get(self):
        return PLAYGROUND_HTML, 200

    def post(self):
        data = request.get_json()
        if data is None or not isinstance(data, dict):
            return 'Bad Request', 400

        variables = data.get('variables')
        if variables and not isinstance(variables, dict):
            return 'Bad Request', 400

        # Note: Passing the request to the context is option. In Flask, the current
        #   request is allways accessible as flask.request.
        result = graphql_sync(
            schema,
            data.get('query'),
            context_value=request,
            variable_values=variables,
            operation_name=data.get('operationName')
        )

        response = {"data": result.data}
        if result.errors:
            response["errors"] = [format_error(e) for e in result.errors]

        return jsonify(response)
````

**Comment by commandtab** [comment]
Does this example respond to GraphQL queries on the `/` route, or `/graphql` (or something else)? If something other than `/`, I'm curious where that route path is defined. Thanks for posting this!

**Comment by NyanKiyoshi** [comment]
Flask user will have to add a rule to their `Flask` app. They would do this:
```python
app.add_url_rule('/graphql/', view_func=GraphQlView.as_view('graphql'))`
```

That would be good to add as note in the documentation, as most Flask users don't use method views, i.e. blueprints or `app.route`!

jrast:
Maybe it would be better to rewrite the example using normal view functions.

rafalp:
It's been a while since I've did anything with flask, but wouldn't something like this be possible?

```python
@app.route('/graphql/', methods=['GET'])
def graphql_playground():
    return PLAYGROUND_HTML, 200


@app.route('/graphql/', methods=['POST'])
def graphql_server():
    ...
```

jrast:
[...]

PR #126, opened 2019-04-01
Added a custom server example using Flask

jrast:
See issue https://github.com/mirumee/ariadne/issues/125

### DIFF
--- a/docs/custom-server.rst
+++ b/docs/custom-server.rst
@@ -102,3 +102,66 @@ The following example presents a basic GraphQL server using a Django framework::
         status_code = 200 if success else 400
         # Send response to client
         return JsonResponse(result, status=status_code)
+
+
+Basic GraphQL server with Flask
+--------------------------------
+
+The following example presents a basic GraphQL server using a Flask::
+
+    from flask import Flask, request, jsonify
+    from ariadne import QueryType, graphql_sync, make_executable_schema
+    from ariadne.constants import PLAYGROUND_HTML
+
+
+    type_defs = """
+        type Query {
+            hello: String!
+        }
+    """
+
+    query = QueryType()
+
+
+    @query.field("hello")
+    def resolve_hello(_, info):
+        request = info.context
+        user_agent = request.headers.get("User-Agent", "Guest")
+        return "Hello, %s!" % user_agent
+
+
+    app = Flask(__name__)
+    schema = make_executable_schema(type_defs, query)
+
+
+    @app.route('/graphql', methods=['GET'])
+    def graphql_playgroud():
+        """Serving the GraphQL Playground
+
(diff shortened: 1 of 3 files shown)


jrast:
Thanks for the review @tbsf 
I will update the PR accordingly.

jrast:
Ok, I updated the example. It's working now with Ariadne 0.3.0 and includes the recomendations of @tbsf

rafalp:
I'm wondering if this shouldn't be 3rd party library, eg. `flask_ariadne` with readme, tests and Pypi release.

I'm not strongly opinionated but I'm worrying that we may be setting an expectation to the folk that our docs will eventually cover all Python web frameworks, and that's not going to work.

tbsf:
This is just my experience, but finding a Flask example was a deciding factor in giving this a try instead of sticking with Apollo. I didn't even notice there was also documentation for Django. Regardless of the route you choose to go down for documentation, I definitely suggest calling more attention to the ability to create a custom servers and mention that Django and Flask are supported in the README itself.

See Apollo's list of integrations for example: https://github.com/apollographql/apollo-server#installation-integrations

rafalp:
@tbsf this is excellent feedback, I haven't considered this. Thanks!

jrast:
I don't see the need for a 3rd party library at the moment. It's only about 20 lines of code which are needed and a package for these lines seems to be an overkill and would remind me of some npm packages.

On the other side, I agree that it will be impossible to have a example for each web framework. And if a lot of framework specific examples are in the docs, issues about the framework integration will pop up which are difficult to solve if the maintainers are not working with these frameworks.
Maybe a note can be added in the docs that these examples are more a "Proof of Concept" and no support will be given?

rafalp:
I'll do the pragmatic thing here and assume I've been trying to optimize prematurely ;)

@jrast can you please update the PR to rename "custom servers" to "Integrations" and rename "Basic GraphQL server with ..." to "... example"? I know @patrys also wants to follow up with "Starlette example".

docs/custom-server.rst
```diff
@@ -102,3 +102,66 @@ The following example presents a basic GraphQL server using a Django framework::
         status_code = 200 if success else 400
         # Send response to client
         return JsonResponse(result, status=status_code)
+
+
+Basic GraphQL server with Flask
```
patrys:
```suggestion
Integrating with Flask
```

"Basic" could be taken to mean "that's not suitable for real-world use".

rafalp: Starting chapters with framework name will make for better copy  and navigation IMHO:

```
- Django integration
- Flask integration
- Starlette integration
```

rafalp:
@jrast thank you for help! I'll close this PR and open my own that includes `Flask` chapter you've written as well some more generic pointers for other integration authors.

PR #131, opened 2019-04-04
Setup black to ignore tests snapshots

rafalp:
This PR adds `pyproject.toml` file that excludes bunch of directories from black, including `snapshots` dir created by pytest snapshot plugin.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,19 @@
+[tool.black]
+line-length = 88
+target-version = ['py36', 'py37', 'py38']
+include = '\.pyi?$'
+exclude = '''
+/(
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+  | \.hg
+  | \.mypy_cache
+  | \.tox
+  | \.venv
+  | _build
+  | buck-out
+  | build
+  | dist
+  | snapshots
+  # The following are specific to Black, you probably don't want those.
```
patrys: Do we want them?

rafalp: We probably don't want those.

PR #136, opened 2019-04-10
Removed duplicate import in Django example

jrast:
The Django example imported `graphql_sync` from ariadne but also from graphql. The import from graphql should be removed and is a artifact of the old example when ariadne offered no graphql_sync function (before version 0.3.0).


rafalp:
Nice catch, thanks!

PR #139, opened 2019-04-12
ASGI 3.0

tomchristie:
See https://www.aeracode.org/2019/03/20/asgi-30/

As of Tuesday, all the latest releases of Uvicorn, Daphne, and Hypercorn now support ASGI 3.
Starlette's master supports ASGI 3, but is currently a pre-release (`pip install starlette --pre`)

I'd suggest we probably want to co-ordinate a Starlette and Ariadne release to both switch to ASGI 3.0 at the same time.

(Yeah, it's a pain, but we figured this was the last possible moment we could still get away with a change like this, and it's worth it for the simplification it brings.)

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -43,12 +43,13 @@ class GraphQL:
         self.keepalive = keepalive
         self.schema = schema
 
-    def __call__(self, scope: Scope):
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
         if scope["type"] == "http":
-            return partial(self.handle_http, scope=scope)
-        if scope["type"] == "websocket":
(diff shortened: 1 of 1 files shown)


rafalp:
Right, if anything we can release `0.4` with just ASGI update.

tomchristie:
Neat, thanks.
You folks planning on rolling a release with this? May push a starlette ASGI 3 update once that’s done.

rafalp:
Hey,

We've actually been holding out with `0.4` release for Starlette ASGI 3 to get out of beta, merging this mostly for dev convenience ATM.

Unless you have anything planned, You don't have to wait for us, ship and we will follow :)

PR #141, opened 2019-04-13
Update documentation of context_for_request method to match definition

ajhyndman:
It looks like this method definition in the docs is slightly out of sync with the implementation. 🙂 

```
  File "lib/python3.7/site-packages/ariadne/asgi.py", line 100, in graphql_http_server
    context_value = await self.context_for_request(request, data)
TypeError: context_for_request() takes 2 positional arguments but 3 were given
```

### DIFF
--- a/docs/asgi.rst
+++ b/docs/asgi.rst
@@ -39,9 +39,10 @@ Customizing context or root
     :param variables: an optional `dict` representing the query variables.
     :return: value that should be passed to root resolvers as the parent (first argument).
 
-.. method:: GraphQL.context_for_request(request)
+.. method:: GraphQL.context_for_request(request, data)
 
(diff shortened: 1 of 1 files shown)


docs/asgi.rst
```diff
@@ -39,9 +39,10 @@ Customizing context or root
     :param variables: an optional `dict` representing the query variables.
     :return: value that should be passed to root resolvers as the parent (first argument).
 
-.. method:: GraphQL.context_for_request(request)
+.. method:: GraphQL.context_for_request(request, data)
 
     :param request: either a `Request` sent by the client or a message sent over a `WebSocket`.
+    :param data: the request's JSON content, if defined
```
rafalp:
```suggestion
    :param data: the request's JSON content, if defined.
```

PR #142, opened 2019-04-13
Ugly first implementation of the tracing extension

patrys:
Implements [Apollo Tracing](https://github.com/apollographql/apollo-tracing) supported by Playground and Apollo Engine.

Tracing is disabled until explicitly allowed. Tracing is only enabled if client explicitly includes `X-Apollo-Tracing` in HTTP headers. Tracing is disabled for the default resolver and for hidden structures (those that start with two underscores) to avoid the huge overhead of tracing an introspection query.


rafalp:
Superseded by #215

PR #143, opened 2019-04-14
Initial attempt at mocking servers

patrys:
Opening this as a draft to start the discussion. This gives you a set of tools to return fake (but compliant) responses for any schema and means to provide custom factories for certain data types.

To play with it:

```python
from ariadne import load_schema_from_path, make_executable_schema
from ariadne.mocking import FactoryMap, MockResolverSetter

type_defs = load_schema_from_path("schema.graphql")

factory_map = FactoryMap()


@factory_map.type("Int")
def fake_int(value):
    # value is whatever the parent's key/attribute
    # with the same name as this field contained
    if value is not None:
        # if parent resolver provided a value, use it
        return value
    return 42


schema = make_executable_schema(type_defs, [..., MockResolverSetter(factory_map)])
app = GraphQL(schema, debug=True)
```

### DIFF
--- a/ariadne/mocking.py
+++ b/ariadne/mocking.py
@@ -1,0 +1,126 @@
+import random
+from functools import partial
+from inspect import isawaitable
+from typing import Any
+
+from ariadne import FallbackResolversSetter
+from graphql.type import (
+    GraphQLEnumType,
+    GraphQLField,
+    GraphQLList,
+    GraphQLNonNull,
+    GraphQLObjectType,
+    GraphQLOutputType,
+    GraphQLResolveInfo,
+    GraphQLScalarType,
+    GraphQLUnionType,
+    ResponsePath,
+)
+
+
+def generate_scalar(value, path, of_type):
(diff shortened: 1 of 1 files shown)


patrys:
CC: @tomchristie as you've briefly mentioned similar services.

salwator:
I was recently thinking about GQL Schema mocking / faking issue and came to some requirements that might be worth considering:
- when on non-production env mixing up _real_ data with _fake_ is convenient to work on features until backend implements anything, cross-referencing might be hard but extremely valuable 
- most useful way to fake is IMO by providing hierarchical data structure, maybe external *json* so it can be easily modified by frontend devs
- integrating something like [faker](https://github.com/joke2k/faker) needs to be relatively simple

patrys:
@salwator I'm using the above with faker. It will respect whichever real resolvers you already have in place so you can add real data progressively as the implementation happens. I specifically did not want to optimize around static data sources as this contributes to a problem of client code optimizing for a particular set of data.

PR #147, opened 2019-04-18
doc: fix ASGI example

tkamenoko:
https://github.com/mirumee/ariadne/blob/ec4a9b9867bc53bda3ed1c1675bf16ff121ddba4/ariadne/asgi.py#L53-L62

These methods must be async, fix example.


rafalp:
Good catch, thank you 😊

PR #150, opened 2019-05-06
Update graphql-core-next to 1.0.3

rafalp:
This PR updates `graphql-core-next` to 1.0.3, giving Ariadne parity with GraphQL.js 14.2.1.


rafalp:
And now our types have exploded because `graphql-core-next` is now coming with improved types definitions.

PR #158, opened 2019-05-07
Set scalar default literal parser when value parser is set

### LINKED ISSUE #26: If parse_literal is not present try fallbacking to `parse_value(ast.value)`
Following idea was brought up in discussion for #24:

> Maybe we could default to calling parse_value with ast.value when only one function is provided?

This requires further study. `IntValue`, `StringValue` and friends are obvious to deal with, but but complex types like `ListValue` may require some extra unpacking magic.

Still, if it is possible to pull off, it could be an excellent convenience for developers creating custom scalars, saving the need for potentially maintaining two very simiiar implementations, one doing `isinstance(value, basestr)` and other `isinstance(value, StringValue)`.

rafalp:
This PR sets default literal parser calling value parser with `ast.value`, saving some boilerplate when writing simple custom scalars.

Fixes #26

### DIFF
--- a/tests/test_custom_scalars.py
+++ b/tests/test_custom_scalars.py
@@ -12,10 +12,12 @@ TEST_DATE_SERIALIZED = TEST_DATE.strftime("%Y-%m-%d")
 type_defs = """
     scalar DateReadOnly
     scalar DateInput
+    scalar ScalarWithDefaultParser
 
     type Query {
         testSerialize: DateReadOnly!
         testInput(value: DateInput!): Boolean!
+        testInputValueType(value: ScalarWithDefaultParser!): String!
     }
 """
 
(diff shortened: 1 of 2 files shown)


tests/test_custom_scalars.py
```diff
@@ -60,7 +67,21 @@ def parse_date_literal(ast):
...
 
-schema = make_executable_schema(type_defs, [query, datereadonly, dateinput])
+scalar_with_default_parser = ScalarType("ScalarWithDefaultParser")
+
+
+@scalar_with_default_parser.value_parser
+def parse_value_from_default_literal_parser(value):
+    print(value)
+    if not isinstance(value, (str, bool)):
+        raise ValueError()
```
levrik: I'm currently trying to figure out what the correct method for checking the type of a custom scalar is.
Looks like either you check the type of the value in a value parser or the type of the AST node in a literal parser.
I ended up here while researching that.
Now I wonder why a `ValueError` is raised here. Shouldn't that be a `TypeError` instead or is there something I don't get? Also, what's the proper way for parsing a date out of a string? Checking the AST node type in a literal parser or checking the value type in a value parser?

rafalp: @levrik We are happy to discuss this, but comment to code fragment in merged PR feels quite wrong place to do this. Could you start discussion on [our Spectrum](https://spectrum.chat/ariadne/general/) or new issue? Thanks!

Issue #95, opened 2019-01-23
Extending a Schema


VDigitall:
Hi, it is very useful library, thanks.
Do you plan to implement support extending a schema, [documentation](https://graphql-core-next.readthedocs.io/en/latest/usage/extension.html)?
In my case are default scheme and extensions. I need to extend scheme by some extensions according some rule.

rafalp:
We want to have support for extending schema in Ariadne, but right now we don't have any ideas on how those should look like. Maybe we should add `extensions=` to `make_executable_schema`?

> I need to extend scheme by some extensions according some rule.

Can you tell us more about your use-case? Do you need to extend schema based on application configuration? Or before query is executed, based on authenticated user?

And thanks!

VDigitall:
In my use-case need to extend schema before query is executed, based on authenticated user.
> Maybe we should add extensions= to make_executable_schema?
This is the first thing that came to mind me, too.

patrys:
@VDigitall As a workaround, would it be possible for you to build two schemas and choose the one to execute the query against based on the user? It does not seem like a good idea to rebuild the schema each time a request is made.

VDigitall:
Yes, it is a good idea, but it will be a good approach, when I can store the base schema in one directory and extensions for these schema in the other, and can mix them, instead describing a new full schema.

VDigitall:
For example in my app:
```
my_schema_1 = make_executable_schema(type_defs, extensions=[ext_1, ext_2], resolvers)
my_schema_2 = make_executable_schema(type_defs, extensions=[ext_2, ext_3], resolvers)
my_schema_3 = make_executable_schema(type_defs, extensions=[ext_1, ext_3], resolvers)
```
also in some cases resolvers can be different, too.

VDigitall:
Yes I see docs on modularization.
I mind when I have:
```
user_defs = """
type Product {
  sku: String
  title: String
  keywords: String
  available: Boolean
  price: Float
}"""
```
and now for admin I need to add field `supplier: Supplier`

In your case it would be like:
```
admin_defs = """
type Product {
  sku: String
  title: String
  keywords: String
  available: Boolean
  price: Float
  supplier: Supplier
}"""
```

but for me fit 

```
product_defs = """
type Product {
  sku: String
  title: String
  keywords: String
  available: Boolean
  price: Float
}"""

product_ext_supplier = """
extend type Product {
  supplier: Supplier
}
"""

product_ext_comments = """
extend type Product {
[...]

PR #166, opened 2019-05-08
Extend schema types

rafalp:
This PR changes `make_executable_schema` implementation so its possible to use `extend type` in passed SDL, opening new ways for applications modularization.

Fixes #135 #95

### DIFF
--- a/ariadne/executable_schema.py
+++ b/ariadne/executable_schema.py
@@ -39,11 +39,12 @@ def build_and_extend_schema(ast: DocumentNode) -> GraphQLSchema:
 
 
 EXTENSION_KINDS = [
+    "scalar_type_extension",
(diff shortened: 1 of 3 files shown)


ariadne/executable_schema.py
```diff
@@ -25,3 +26,27 @@ def make_executable_schema(
...
+
+    return schema
+
+
+EXTENSION_KINDS = [
+    "object_type_extension",
+    "interface_type_extension",
+    "input_object_type_extension",
+    "union_type_extension",
+    "enum_type_extension",
```
patrys: It seems we're missing `scalar_type_extension`?

rafalp: I've updated the implementation to also include scalar extension.

PR #170, opened 2019-05-10
Document integration with Starlette

patrys:
Integration is fairly straightforward with mounts. Also unifies quotes in the Flask example.


rafalp:
Can you update changelog too?

PR #185, opened 2019-05-24
Add a query cost validator

rafalp:
Ported code from https://github.com/pa-bru/graphql-cost-analysis

This PR continues the work started by @patrys on adding query cost validator to Ariadne's library.

### TODO

* [x] Add `ariadne.validation` package
* [x] Validate that cost map doesn't define non-existing fields
* [x] Update `graphql()` to support callable `validation_rules` for delayed/conditional validation initialization
* [x] Improve tests coverage
* [x] Add type annotations to query cost validator
* [x] Update changelog


manesioz:
@rafalp hey! need any help with anything here?

rafalp:
Added one more test to websocket handling in ASGI, so now we have three:

- when client's payload is `{"query": "subscription { field }"}`
- when client's payload is `{"query": "subscription { field }", "operationName": null}`
- when client's payload is `{"query": "subscription OperationName { field }", "operationName": "OperationName"}`

Each of those is valid payload sent by GraphQL clients.

rafalp:
Marking this for 0.12 release, so it won't block 0.11 release.

mavaras:
Hi! I have a question, what's the way to use that cost_validator() function?
I guess including it in validation_rules, but in which way?
Thanks everyone

rafalp:
Correct, you pass it to GraphQL app via `validation_rules=[cost_validator(...)]`.

mavaras:
Importing it with `from ariadne.validation import cost_validator` returns me an error 'no validation module in Ariadne'.
What can I do? My Ariadne version is 0.11

rafalp:
Please note that this feature wasn't released yet. Otherwise there would've been a documentation for it. :)

PR #194, opened 2019-06-05
CORS support

bogdal:
This PR allows users to enable CORS.

An example (_urls.py_):
```python
path('graphql/', GraphQLView.as_view(allowed_origin='https://example.com', schema=schema, ), name='graphql'),
```

### DIFF
--- a/ariadne/contrib/django/views.py
+++ b/ariadne/contrib/django/views.py
@@ -30,6 +30,7 @@ class GraphQLView(TemplateView):
     validation_rules = None
     error_formatter: Optional[ErrorFormatter] = None
     middleware = None
+    allowed_origin = None
 
     def get(
         self, request: HttpRequest, *args, **kwargs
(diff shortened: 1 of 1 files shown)


patrys:
I'd say this is not a concern of Ariadne. There's a very popular package you can use (and we should probably document): https://github.com/ottoyiu/django-cors-headers

bogdal:
My motivation was to provide a complete Django's view which allows you to work with GraphQL API out of the box. But I understand your point of view. For sure we should mention about the above package in the doc to help users build working API.

patrys:
Well, it does work by default as in it accepts queries and executes them.

Exposing HTTP endpoints over CORS is a separate problem from executing GraphQL queries. Starlette offers similar separation where you use middleware to manage CORS: https://www.starlette.io/middleware/#corsmiddleware

Issue #200, opened 2019-06-10
Should EnumType accept internal values when using enum.Enum?


ajhyndman:
This is a little subtle, so let me see if I can make it clear enough to get the point across.

Currently, if you use `EnumType` and pass an `Enum` class as the value mapping, the behaviour isn't exactly what I expected.

I have tried creating an enum class that maps from GraphQL enum names to the internal string values used by our server.

```python
class TeamType(enum.Enum):
  ORG = 'org'
  MANAGER = 'manager'

team_type = EnumType('TeamType', TeamType)
```

By reading Ariadne's implementation, I can see that one of my resolvers receives a value which is an instance of Enum value (i.e. `TeamType .ORG`), then Ariadne will correctly transform that and return the associated enum value's name.  However, if my resolver receives the actual _string_ value we used when defining the `TeamType` enum, then Ariadne does not recognize that value, and will throw a GraphQLError.

Of course, this can be worked around trivially, using a dictionary instead of `Enum`.

```python
team_type = EnumType('TeamType', {'ORG': 'org', 'MANAGER': 'manager'})
```

But it feels like a bit of a rough edge to me.  Or it might limit the utility of supporting Enum to define transformation.  Thoughts?

rafalp:
Looks like a BUG to me. `{ myField(enum: ORG) }` should pass `TeamType.ORG` to the python resolver, and resolver returning the `TeamType.ORG` should result in `ORG` appearing in the query result.

ajhyndman:
@rafalp To clarify:

If a resolver returns `TeamType.ORG`, then `'ORG'` correctly appears in the query result.

What I wasn't expecting (but maybe is okay?) was that if the resolver returns `'org'`, then the resolver throws a validation error, instead of `'ORG'` appearing in the query result.

rafalp:
So long story short, `Enum.ORG` is not equal to `org`. You need to update your enum definition:

```
class TeamType(str, enum.Enum):
  ORG = 'org'
  MANAGER = 'manager'
```

That will fix it.

ajhyndman:
@rafalp Oh cool.

That definitely solves my use case.  Thank you for investigating!

PR #201, opened 2019-06-10
Should EnumType accept internal values when using enum.Enum?

jose-humu:
Resolving following issue: https://github.com/mirumee/ariadne/issues/200

### DIFF
--- a/tests/test_enums.py
+++ b/tests/test_enums.py
@@ -185,3 +185,33 @@ def test_int_enum_arg_is_transformed_to_internal_value():
 
     result = graphql_sync(schema, "{ testEnum(value: EMPIRE) }")
     assert result.data["testEnum"] is False
+
+
+class PyStringEnum(Enum):
+    NEWHOPE = "1977"
+    EMPIRE = "1980"
+    JEDI = "1983"
+
+
+py_string_enum = EnumType("Episode", PyStringEnum)
+
+
+def test_enum_is_resolved_from_internal_value():
+    query = QueryType()
+    query.set_field("testEnum", lambda *_: PyStringEnum.NEWHOPE)
+
+    schema = make_executable_schema([enum_definition, enum_field], [query, py_string_enum])
+    result = graphql_sync(schema, "{ testEnum }")
+    assert result.data["testEnum"] == TEST_VALUE
+
+
+def test_enum_arg_is_transformed_to_internal_value():
+    query = QueryType()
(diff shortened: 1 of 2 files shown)


tests/test_enums.py
```diff
@@ -185,3 +185,33 @@ def test_int_enum_arg_is_transformed_to_internal_value():
...
+    query.set_field("testEnum", lambda *_: PyStringEnum.NEWHOPE)
+
+    schema = make_executable_schema([enum_definition, enum_field], [query, py_string_enum])
+    result = graphql_sync(schema, "{ testEnum }")
+    assert result.data["testEnum"] == TEST_VALUE
+
+
+def test_enum_string_arg_is_transformed_to_internal_value():
+    query = QueryType()
+    query.set_field("testEnum", lambda *_, value: value == PyStringEnum.NEWHOPE)
```
patrys: You should be able to use identity here instead of equality, enums are singletons.

tests/test_enums.py
```diff
@@ -185,3 +185,33 @@ def test_int_enum_arg_is_transformed_to_internal_value():
...
+py_string_enum = EnumType("Episode", PyStringEnum)
+
+
+def test_enum_string__is_resolved_from_internal_value():
+    query = QueryType()
+    query.set_field("testEnum", lambda *_: PyStringEnum.NEWHOPE)
+
+    schema = make_executable_schema([enum_definition, enum_field], [query, py_string_enum])
+    result = graphql_sync(schema, "{ testEnum }")
+    assert result.data["testEnum"] == TEST_VALUE
```
patrys: Could we compare this against `PyStringEnum.NEWHOPE`? I hope to kill `TEST_VALUE` one day as it makes tests unreadable.

tests/test_enums.py
```diff
@@ -185,3 +185,33 @@ def test_int_enum_arg_is_transformed_to_internal_value():
...
+class PyStringEnum(Enum):
+    NEWHOPE = "1977"
+    EMPIRE = "1980"
+    JEDI = "1983"
+
+
+py_string_enum = EnumType("Episode", PyStringEnum)
+
+
+def test_enum_string__is_resolved_from_internal_value():
```
patrys: Typo: double underscore

rafalp:
Thank you for your PR!  We've investigated this and found a solution that doesn't require us for modifying our `EnumType` implementation. <3

PR #202, opened 2019-06-12
Add extensions support

rafalp:
This PR adds extension system inspired by one implemented in Apollo Server.

I've hoped for this system to let real extensibility in Ariadne, but it seems that it was designed mostly for implementing monitoring, and little else. - Our hands are also tied by whole `sync`/`async` stuff and we really wanted to avoid having separate extension system for sync and async execution.


ariadne/graphql.py
```diff
@@ -17,48 +37,73 @@
...
-        )
-    except GraphQLError as error:
-        return handle_graphql_errors(
-            [error], logger=logger, error_formatter=error_formatter, debug=debug
-        )
-    else:
-        return handle_query_result(
-            result, logger=logger, error_formatter=error_formatter, debug=debug
-        )
+            with extension_manager.validation(context_value):
```
patrys: Should this be `validating(...)` (to match `parsing(...)`)?

ariadne/graphql.py
```diff
@@ -17,48 +37,73 @@
...
+                if validation_errors:
+                    return handle_graphql_errors(
+                        validation_errors,
+                        logger=logger,
+                        error_formatter=error_formatter,
+                        debug=debug,
+                        extension_manager=extension_manager,
+                    )
+
+            with extension_manager.execution(context_value):
```
patrys: Should this be `executing(...)` (see above)?

ariadne/graphql.py
rafalp: Those are named after [`graphql-extensions`](https://github.com/apollographql/apollo-server/blob/master/packages/graphql-extensions/src/index.ts#L102) which does `parsing`, `validation` and `execution`, but seeing how this API's not publish in Apollo Server, we may go for something else. I am not sure if I like `def validating` call `validation_started` & `validation_finished`

PR #205, opened 2019-06-25
Not send/recv websocket messages after a client disconnects

ciscorn:
The following error occurs if a browser closes/reloads during a subscription resolver is running.

```
ERROR: Task exception was never retrieved
future: <Task finished coro=<GraphQL.observe_async_results() done, defined at ./ariadne/asgi.py:229> exception=RuntimeError("Unexpected ASGI message 'websocket.send', after sending 'websocket.close'.")>
Traceback (most recent call last):
  File "./ariadne/asgi.py", line 257, in observe_async_results
    await websocket.send_json({"type": GQL_COMPLETE, "id": operation_id})
  File "/usr/local/lib/python3.7/site-packages/starlette/websockets.py", line 116, in send_json
    await self.send({"type": "websocket.send", "text": text})
  File "/usr/local/lib/python3.7/site-packages/starlette/websockets.py", line 68, in send
    await self._send(message)
  File "./myapp/graphql_test.py", line 57, in sender
    await send(message)
  File "/usr/local/lib/python3.7/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 217, in asgi_send
    raise RuntimeError(msg % message_type)
RuntimeError: Unexpected ASGI message 'websocket.send', after sending 'websocket.close'.
```

We should check not only `websocket.application_state` but also `websocket.client_state` when sending `GQL_COMPLETE`

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -155,7 +155,10 @@ class GraphQL:
         subscriptions: Dict[str, AsyncGenerator] = {}
         await websocket.accept("graphql-ws")
         try:
-            while websocket.application_state != WebSocketState.DISCONNECTED:
+            while (
+                websocket.client_state != WebSocketState.DISCONNECTED
+                and websocket.application_state != WebSocketState.DISCONNECTED
+            ):
                 message = await websocket.receive_json()
                 await self.handle_websocket_message(message, websocket, subscriptions)
         except WebSocketDisconnect:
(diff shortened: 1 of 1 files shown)


ariadne/asgi.py
```diff
@@ -253,5 +253,8 @@ def __init__(
                 {"type": GQL_DATA, "id": operation_id, "payload": payload}
             )
 
-        if websocket.application_state != WebSocketState.DISCONNECTED:
+        if (
```
patrys: Do we need to check both conditions or does client disconnecting imply application state being disconnected as well?

ciscorn: I think we need check both.

When a client sends `GQL_CONNECTION_TERMINATE`, server's `handle_websocket_message(...)` calls `websocket.close()` which immediately change the application_state DISCONNECTED (but the client_state is still CONNECTED)

rafalp:
Thanks, I'll merge this later today and update changelog.

rafalp:
Well, here's my "today", 27 days later.

Merged via #225, thank you!

PR #215, opened 2019-07-19
Add tracing extensions

### LINKED ISSUE #173: Look into supporting OpenTracing
OpenTracing (https://opentracing.io/) looks like a great way to implement generic APM support.

OpenTracing would also allow us to track requests across multiple servers (via forwarding trace/span IDs, for example using HTTP headers).

rafalp:
This PR adds Apollo Tracing and OpenTracing extensions to Ariadne. Part of this issue also simplifies the extension system because we've started out big, and we overdid it a little with number of hooks.

### TODO

* [x] Add Apollo Tracing extension
* [x] Add OpenTracing extension
* [x] Write tests
* [x] Add to changelog
* [x] Add to features list in readme and docs index

Fixes #173


ariadne/contrib/tracing/utils.py
```diff
@@ -0,0 +1,35 @@
...
+        path = path.prev
+    return elements[::-1]
+
+
+def should_trace(info: GraphQLResolveInfo):
+    if info.field_name not in info.parent_type.fields:
+        return False
+
+    resolver = info.parent_type.fields[info.field_name].resolve
+    if (
```
salwator: We could just return expression value (inversed).

ariadne/contrib/tracing/utils.py
```diff
@@ -0,0 +1,35 @@
...
+        or is_introspection_field(info)
+    ):
+        return False
+
+    return True
+
+
+def is_introspection_field(info: GraphQLResolveInfo):
+    path = info.path
+    while path:
```
salwator: I would reconsider refactoring it to use _any_.

tests/tracing/test_apollotracing.py
```diff
@@ -0,0 +1,70 @@
...
+    schema
+):
+    _, result = await graphql(
+        schema, {"query": "{ status }"}, extensions=[ApolloTracingExtension]
+    )
+
+    resolvers = result["extensions"]["tracing"]["execution"]["resolvers"]
+
+    assert len(resolvers) == 1
+    assert resolvers[0]["path"] == ["status"]
```
salwator: Maybe we can use snapshots in those tests?
If there is problem with time and duration results we could mock _time_ fetching or use freezegun.

ariadne/contrib/tracing/utils.py
rafalp: Nice catch!

PR #216, opened 2019-07-19
Update graphql-core-next to 1.1.1

rafalp:
This PR updates `graphql-core-next` to 1.1.1, fixing our tests failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.6.0 (unreleased)
 
+- Updated `graphql-core-next` to 1.1.1 which has feature parity with GraphQL.js v14.4.0.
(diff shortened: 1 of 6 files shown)


hvdklauw:
Also don't see the actual update of graphql-core-next in the changeset.

rafalp:
@hvdklauw we don't really freeze `graphql-core-next` version to any specific number, and its maintainer has announced that next major release will happen as `graphql-core` package.

hvdklauw:
You are right for the setup.py, but the [requirements.txt](https://github.com/mirumee/ariadne/blob/master/requirements.txt) actually does have a frozen version.

rafalp:
Nice catch, I'll fix that one too 👍

PR #221, opened 2019-07-26
Don't require typing-extensions for python3.7

tbsf:
Remove unnecessary typing-extensions requirement for python3.7 and up.


tbsf:
I found Protocol isn't included until Python3.8. I bumped the version check to 3.8, but it's still early.

PR #167, opened 2019-05-08
First working custom directive implementation

rafalp:
This PR adds `DirectiveType` to Ariadne that allows developers to define custom field directive that can override field attributes depending on its own:

```
from ariadne import DirectiveType

block_user_agent = DirectiveType("blockUserAgent")

@block_user_agent.field
def field_visitor(field, *, agent):
    original_resolver = field.resolve
    def resolve_user_agent_check(obj, info, **kwargs):
        if agent.lower() not in info.context["request"].headers["user-agent"].lower():
            return original_resolver(obj, info, **kwargs)
        return None

    field.resolve = resolve_user_agent_check
```

Fixes #138

Issue #138, opened 2019-04-12
Support passing custom directives to `make_executable_schema`


rafalp:
[GraphQL spec](https://graphql.github.io/graphql-spec/June2018/#sec-Language.Directives) describes directives as a way to describe alternate runtime execution and type validation behavior in a GraphQL document.

~~After quick look around `make_executable_schema` could take `*, schema_directives=List[Directive]` kwarg with list of callables implementing GraphQL directives that should be passed to `graphql-core-next` when an executable schema is created from SDL.~~

We've discussed this internally and idea is to add special `DirectiveType` that would visit appropriate parts of `GraphQLSchema` in it's `bind_to_schema`, calling user-provided functions on different elements. For example, `permission` field directive could look like this in Ariadne:

```
from ariadne import DirectiveType

with_permission = DirectiveType("withPermission")

@with_permission.field:
def field_visitor(field, info, *, permission):
    # field is an instance of GraphQLField in schema that has `withPermission` directive
    original_resolver = field.resolve
    def resolve_with_permission_check(obj, info, **kwargs):
        if info.context.user.has_permission(permission):
            return original_resolver(obj, info, **kwargs)
        return None

    field.resolve = resolve_with_permission_check
```

This directive would then allow schema designers to protect fields in schema:

```
directive @withPermission(
  permission: String
) on FIELD_DEFINITION

type User {
  ipAddress: String @withPermission(permission: "is_admin")
}
```

rafalp:
Adding this to 0.4.0 milestone because we have some ideas on how to go about implementing those.

prashanthellina:
@rafalp +1 for this support. I am implementing multiple services using `ariadne` (which I picked over `Graphene` because I prefer schema-first approach) where I have to specify custom directives and have the resolvers apply directive specific logic.

I see that this has been removed from 0.4 milestone and isn't part of the 0.6 milestone. When does your team plan to address this issue?

In the meantime, is there a workaround that we can use? i.e. Is there any way to access the applied directive information from inside the resolver method?

rafalp:
We are currently focused on adding monitoring capabilities to Ariadne - we need those for our projects at Mirumee. This is why directives got put on the back-burner.

> In the meantime, is there a workaround that we can use? i.e. Is there any way to access the applied directive information from inside the resolver method?

[...]

PR #222, opened 2019-07-29
Feature - Schema directives visitor

0x1EE7:
@rafalp Here is a the first draft for porting the SchemaDirectiveVisitor from apollo into ariadne as we have discussed in #167 and related to #138 

While the simple test cases provided with this PR is now passing there is some TODO items that needs to be discussed and resolved. I will add some more test cases and also looking forward to your feedback.

### TODOs
- [x] Check if `update_each_key` can only accept a dict. `graphql-core` seems to be always returning dicts
- [x]  Needs improvements to remove nodes in `call_method`. Since python does not have 'undefined', False value is chosen for removal. Still more clarification is needed. Maybe return a tuple instead? But that would have an effect on the visitor methods' signature too.
- [x] How would `heal_schema` be needed?


ariadne/directives.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class DirectiveType(SchemaBindable):
+    def __init__(self, name: str, visitor: Type[SchemaDirectiveVisitor]) -> None:
+        self.name = name
+        self.visitor = visitor
+
+    def bind_to_schema(self, schema: GraphQLSchema) -> None:
+        SchemaDirectiveVisitor.visit_schema_directives(
+            schema, {self.name: self.visitor}
```
rafalp: Could we add `directives: Dict[Str, Type[SchemaDirectiveVisitor]] = None` argument to the `make_executable_schema` instead? That would give us the safety of running directives only after all other bindables set their resolvers/etc/ect.

tests/test_directives.py
```diff
@@ -0,0 +1,176 @@
...
+
+        def resolve_reverse(obj, info, **kwargs):
+            result = original_resolver(obj, info, **kwargs)
+            return result[::-1]
+
+        field.resolve = resolve_reverse
+        return field
+
+
+def test_single_directive_without_args():
```
rafalp: Change test names to have the name of tested logic and result state: `test_field_definition_directive_replaces_field_resolver_with_custom_one`.

rafalp:
> Needs improvements to remove nodes in call_method. Since python does not have 'undefined', False value is chosen for removal.

`False` is good replacement IMHO. Alternatively, we could pack than into a more descriptive constant (`REMOVED_TYPE = False`) or create an empty type like `class RemovedType` and test for instance/type equality, and check for that instead., but thats also going too magical for me.

> How would heal_schema be needed?

From my understanding Apollo's `healSchema` iterates schema types (and their fields) to make sure that following remains true after directives have been run:

```
GraphQLSchema({"User": GraphQLNamedType("User"))
```

Because directives can change schema structure, they can produce broken schema like this:

```
GraphQLSchema({"User": GraphQLNamedType("Author"))
```

So yes, we will need to eventually implement `heal_schema` too.

0x1EE7:
That's great feedback @rafalp I will look into `heal_schema` too.

ariadne/schema_visitor.py
```diff
@@ -149,7 +151,7 @@ def visit_schema(
     ],
 ) -> GraphQLSchema:
     """
-    Helper function that calls visitor_selector and applies the resulting
+    Helper def that calls visitor_selector and applies the resulting
```
rafalp: "function" was better here ;)

0x1EE7: Haha definitely :)

ariadne/executable_schema.py
```diff
@@ -1,21 +1,23 @@
...
 
 from .enums import set_default_enum_values_on_schema
+from .schema_visitor import SchemaDirectiveVisitor
 from .types import SchemaBindable
 
 
 def make_executable_schema(
     type_defs: Union[str, List[str]],
     bindables: Union[SchemaBindable, List[SchemaBindable], None] = None,
+    directives: Dict[str, Type[SchemaDirectiveVisitor]] = None,
```
rafalp: Let's make directives option explicit:

```suggestion
    *,
    directives: Dict[str, Type[SchemaDirectiveVisitor]] = None,
```

0x1EE7:
@rafalp Is there anything more I could do to move this forward?

rafalp:
The plan currently is to release Ariadne 0.6 as soon as possible and then focus releasing custom directives work from this PR as Ariadne 0.7.

rafalp:
Ariadne 0.6 is out, so I'll clone this PR locally sometime next week and test it out myself to see how it works. 🥽

0x1EE7:
> Ariadne 0.6 is out, so I'll clone this PR locally sometime next week and test it out myself to see how it works. 🥽

Great, looking forward to it! 🙇

rafalp:
I've did some quick experiments with this and it works great, but I've noticed you have pushed bunch of fixes yesterday. How do you feel about readyness of this for merging?

This feature is gigantozauric, so its gonna be fun to write reasonable tests harness and docs for it 😅I would like to merge it and then follow up with improvements and tests in next PRs. Perhaps bump version number to 0.7 and release few alphas in the meantime so others can also play with it.

0x1EE7:
> I've did some quick experiments with this and it works great, but I've noticed you have pushed bunch of fixes yesterday. How do you feel about readyness of this for merging?
> 
> This feature is gigantozauric, so its gonna be fun to write reasonable tests harness and docs for it 😅I would like to merge it and then follow up with improvements and tests in next PRs. Perhaps bump version number to 0.7 and release few alphas in the meantime so others can also play with it.

I think that's a great plan. The changes I pushed yesterday was very minimal in order to fix some failures that I noticed while developing some directives on top of this PR. I think it is ready from that point of view and more PRs could follow after this one is merged. 🌟

rafalp:
I've released Ariadne `0.7.0a1`with this PR merged so if anyone wants to give this a try, they can install `ariadne==0.7.0a1`

PR #223, opened 2019-07-31
Fixed ASGI forwarding middleware to graphql executor

nikordaris:
The ASGI app is currently not doing anything with the middleware kwarg. I fixed it to pass it to the `graphql` execution in the `graphql_http_server` function.

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -13,7 +13,7 @@ from typing import (
 )
 
 from graphql import GraphQLError, GraphQLSchema
-from graphql.execution import Middleware
+from graphql.execution import MiddlewareManager
 from starlette.datastructures import UploadFile
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, JSONResponse, PlainTextResponse, Response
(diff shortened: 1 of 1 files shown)


rafalp:
Hi! Thanks for taking the time to contribute. Can you provide tests for this change?

nikordaris:
Lower level tests should already cover this. Code coverage was unchanged. Do you have recommendations for what should be tested here that isn't already covered?

rafalp:
A test should prove that if the middleware is set in this argument, it will be used during query execution.

nikordaris:
To be honest I'm just trying to get a poc comparison going to compare against graphene. Whoever originally added middleware to the asgi function should have tested this because they would have found this bug to begin with. I'm just submitting a one line bug fix and don't have time right now to invest more.

rafalp:
That's fair enough, I'll look at adding test and merging this myself. Thanks for taking time to submit PR!

nikordaris:
Thanks for the quick response though! This lib looks really promising. We have a lot of code in graphene already though so I need to make sure it will be worth the investment to switch over.

rafalp:
I've created PR #230 that includes this commit but also adds tests and alternative ways for setting middleware based on request/context.

Thanks!

PR #228, opened 2019-08-05
Await callable context value in ASGI

### LINKED ISSUE #227: Async context_value callable is not supported
Our ASGI GraphQL app supports callable `context_value`, but it doesnt support `async` callable, so it's impossible to do any async work in it like get user form database. 😂

Example:

```python
async def resolve_context(request: Request) -> Dict:
    return {"user": await get_user_for_request(request)}


graphql_app = GraphQL(schema, context_value=resolve_context)
```

rafalp:
This PR awaits custom callable `context_value`, if its coroutine.

Fixes #227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Resolver returned by `resolve_to` util follows `graphql.default_field_resolver` behaviour and supports resolving to callables.
 - Added `is_default_resolver` utility for checking if resolver function is `graphql.default_field_resolver`, resolver created with `resolve_to` or `alias`.
 - Updated ASGI app disconnection handler to also check client connection state.
+- Fixed `context_value` option on ASGI app not supporting async callables.
 
 
 ## 0.5.0 (2019-06-07)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Resolver returned by `resolve_to` util follows `graphql.default_field_resolver` behaviour and supports resolving to callables.
 - Added `is_default_resolver` utility for checking if resolver function is `graphql.default_field_resolver`, resolver created with `resolve_to` or `alias`.
 - Updated ASGI app disconnection handler to also check client connection state.
+- Fixed `context_value` option on ASGI app not supporting async callables.
```
salwator:
```suggestion
- Fixed ASGI app `context_value` option support for async callables.
```

ariadne/asgi.py
```diff
@@ -79,7 +80,11 @@ def __init__(
 
     async def get_context_for_request(self, request: Any) -> Any:
         if callable(self.context_value):
-            return self.context_value(request)
+            context = self.context_value(request)
+            if isawaitable(context):
```
salwator: We could make it flat, guard-check for _Awaitable_ before this _Callable_ *if* statement. No real reason to keep them together.

rafalp: You can't do it like that - you need to call a function to get coroutine object that you can then test with `isawaitable` and then send to event loop via `await` to get its value.

PR #230, opened 2019-08-08
Sync middleware implementation with extensions

rafalp:
This PR includes #223 but also synces `middleware` option behavior with `extensions`. It also adds tests and changelog.

### DIFF
--- a/ariadne/wsgi.py
+++ b/ariadne/wsgi.py
@@ -1,6 +1,6 @@
 import json
 from cgi import FieldStorage
-from typing import Any, Callable, List, Optional
+from typing import Any, Callable, List, Optional, Union
 
 from graphql import GraphQLError, GraphQLSchema
 from graphql.execution import Middleware, MiddlewareManager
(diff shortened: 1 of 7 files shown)


tests/wsgi/test_configuration.py
```diff
@@ -102,3 +102,23 @@ def test_error_formatter_is_called_with_debug_disabled_flag(schema):
...
+
+
+def test_middlewares_are_passed_to_query_executor(schema):
+    app = GraphQL(schema, middleware=[middleware])
+    _, result = app.execute_query({}, {"query": '{ hello(name: "BOB") }'})
+    assert result == {"data": {"hello": "**Hello, BOB!**"}}
+
+
+def test_middleware_function_result_is_passed_to_query_executor(schema):
+    def get_middleware(*_):
```
salwator: maybe just use lambda inline?

ariadne/wsgi.py
```diff
@@ -149,22 +154,35 @@ def extract_data_from_multipart_request(self, environ: dict) -> Any:
...
 
+    def get_middleware_for_request(
+        self, environ: dict, context: Optional[ContextValue]
+    ) -> Optional[MiddlewareManager]:
+        middleware = self.middleware
+        if callable(middleware):
+            middleware = middleware(environ, context)
+        if middleware:
+            return MiddlewareManager(*middleware)
+        return None
```
salwator: Not really necessary.

rafalp: It's either that or `pylint: disable=...`, so I'll rather `return None` as former will remove useful safety check from pylint.

PR #239, opened 2019-08-21
Support extensions for graphql_sync

tbsf:
(This is a work in progress.)

These changes seemed to work with the following test:

```
from ariadne.types import ExtensionSync
...
class TimestampExtension(ExtensionSync):
    def format(self):
        return {"timestamp": datetime.now().isoformat()}
...
graphql_sync(
    schema,
    ...
    extensions=[TimestampExtension]
```

This also worked if I left off extensions as a kwarg in graphql_sync's call to execute so there may be other test cases that should be checked with this. Is this sufficient to support extensions for synchronous calls or is there another issue that needs to be handled in order for this to work correctly?

### DIFF
--- a/ariadne/graphql.py
+++ b/ariadne/graphql.py
@@ -23,7 +23,6 @@ from .logger import log_error
 from .types import (
     ErrorFormatter,
     Extension,
-    ExtensionSync,
     GraphQLResult,
     RootValue,
     SubscriptionResult,
@@ -115,7 +114,7 @@ def graphql_sync(
     validation_rules: Optional[Sequence[RuleType]] = None,
     error_formatter: ErrorFormatter = format_error,
     middleware: Optional[MiddlewareManager] = None,
-    extensions: Optional[List[Type[ExtensionSync]]] = None,
+    extensions: Optional[List[Type[Extension]]] = None,
     **kwargs,
 ) -> GraphQLResult:
     extension_manager = ExtensionManager(extensions)
--- a/ariadne/graphql.py
+++ b/ariadne/graphql.py
@@ -118,6 +118,7 @@ def graphql_sync(
     **kwargs,
 ) -> GraphQLResult:
     extension_manager = ExtensionManager(extensions)
+
     with extension_manager.request(context_value):
         try:
             validate_data(data)
--- a/ariadne/graphql.py
+++ b/ariadne/graphql.py
@@ -114,59 +114,73 @@ def graphql_sync(
     validation_rules: Optional[Sequence[RuleType]] = None,
     error_formatter: ErrorFormatter = format_error,
     middleware: Optional[MiddlewareManager] = None,
+    extensions: Optional[List[Type[Extension]]] = None,
     **kwargs,
 ) -> GraphQLResult:
-    try:
-        validate_data(data)
-        query, variables, operation_name = (
-            data["query"],
-            data.get("variables"),
-            data.get("operationName"),
-        )
+    extension_manager = ExtensionManager(extensions)
+    with extension_manager.request(context_value):
+        try:
+            validate_data(data)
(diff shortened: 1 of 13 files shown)


rafalp:
Looking good so far! I guess PR should also include the same tests harness that async extensions do. We will likely also need to update the docs for this, and also make a decision, what do we want with extensions we already provide as part of `ariadne.contrib` that are currently async only.

tbsf:
@rafalp I updated the tests I found using extensions with sync versions. I also updated the `ariadne.contrib` extensions with sync versions. Hopefully this is the direction you want to go in. 

I also added werkzeug for the tests I created and extended the client to match what is returned by starlette. You should be able to use this pattern for other tests instead of middleware, but I wasn't sure if these were equivalent.

tbsf:
@rafalp Is there anything else that needs to be done for this to be merged in for 0.0.7? I have added tests and documentation.

rafalp:
@tbsf Sorry about not coming back to you 😅We are now hyper-occupied with other things so work on 0.7 got put on backburner, but I will try to do a review of this sometime this week.

ariadne/graphql.py
```diff
@@ -114,59 +114,73 @@ def graphql_sync(
...
 ) -> GraphQLResult:
-    try:
-        validate_data(data)
-        query, variables, operation_name = (
-            data["query"],
-            data.get("variables"),
-            data.get("operationName"),
-        )
+    extension_manager = ExtensionManager(extensions)
+    with extension_manager.request(context_value):
```
rafalp: In `graphql` there's an empty line between extension_manager and context: 

```suggestion

    with extension_manager.request(context_value):
```

ariadne/types.py
```diff
@@ -49,6 +49,12 @@ def format(self) -> dict:
         return {}  # pragma: no cover
 
 
+class ExtensionSync(Extension):
+    def resolve(self, next_: Resolver, parent: Any, info: GraphQLResolveInfo, **kwargs):
+        result = next_(parent, info, **kwargs)
```
rafalp:
```suggestion
        return next_(parent, info, **kwargs)
```

ariadne/types.py
```diff
@@ -49,6 +49,12 @@ def format(self) -> dict:
...
+        return result
```
rafalp:
```suggestion
```

ariadne/wsgi.py
```diff
@@ -173,6 +181,14 @@ def get_context_for_request(self, environ: dict) -> Optional[ContextValue]:
             return self.context_value(environ)
         return self.context_value or environ
 
+    def get_extensions_for_request(
+        self, environ: dict, context: Optional[ContextValue]
+    ) -> ExtensionList:
+        if callable(self.extensions):
+            extensions = self.extensions(environ, context)
```
rafalp:
```suggestion
            return self.extensions(environ, context)
```

ariadne/wsgi.py
```diff
@@ -173,6 +181,14 @@ def get_context_for_request(self, environ: dict) -> Optional[ContextValue]:
...
+            return extensions
```
rafalp:
```suggestion
```

tbsf:
@rafalp I've updated with your suggested changes. Any suggestions how to fix the code coverage issues?

rafalp:
There's not much that can be done here. Creating extension that does nothing on its own (but inherits `ExtensionSync`) and running query with it could bump the coverage somewhat, and give us harness for default `resolve` implementation for sync extensions.

Besides adding such test, I wouldn't obsess about codecov that much ;)

rafalp:
@tbsf Thanks! I'll also merge docs updates, but I still need to write docs for directives before releasing 0.7.

PR #245, opened 2019-09-12
fixed healing GraphQLUnionType

whiskybar:
Implementing my first GQL directive, I came across an error... it seemed like a typo when I looked around the code. It worked with this, feel free to discard if it needs more attention.


rafalp:
Thanks for taking time to contribute a fix! 👍

PR #247, opened 2019-09-18
Iterate over copy of the map in case directives need to add types

0x1EE7:
This might be handy and make it easier to implement some directives like `@key` that needs to create a `_Entity` union type etc.

### DIFF
--- a/tests/test_directives.py
+++ b/tests/test_directives.py
@@ -13,6 +13,7 @@ from graphql.type import (
     GraphQLInterfaceType,
     GraphQLObjectType,
     GraphQLSchema,
+    GraphQLUnionType,
 )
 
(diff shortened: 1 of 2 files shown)


rafalp:
@0x1EE7 Could you also contribute test illustrating this use as part of this PR? I feel that custom directives cover so many use-cases, any little test added for it helps 👍

tests/test_directives.py
```diff
@@ -507,3 +508,32 @@ def _check_results(result, *, data=None, errors=None):
         exec_with_role("ADMIN"),
         data={"users": [{"name": "Ben", "banned": True, "canPost": False}]},
     )
+
+
+def test_directive_can_add_new_type():
```
rafalp:
```suggestion
def test_directive_can_add_new_type_to_schema():
```

PR #260, opened 2019-11-07
Recursively load schema from path

salwator:
Now, when providing path to load multiple schema files, we traverse the tree in search of _.graphql_ files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 0.8.0 (UNRELEASED)
+
+- Added recursive loading of GraphQL schema files from provided path.
+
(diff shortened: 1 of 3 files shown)


patrys:
I woud love to remove this completely (so the function does not have to guess whether you gave it a schema or a file name) and instead add `load_schema_from_glob("foo/**/*.graphql")`.

rafalp:
@salwator could you also update `CHANGELOG` to include `0.8 (UNRELEASED)` with this change?

rafalp:
@patrys yup, I'll open tracking issue for that later so we can go around implementing it sometime in the future.

This PR just makes our API do what docs are promising they do.

PR #264, opened 2019-11-14
Support for @defer query directive

patrys:
Proof of concept: missing docs, tests, and error support

### DIFF
--- a/ariadne/graphql.py
+++ b/ariadne/graphql.py
@@ -1,22 +1,27 @@
 from asyncio import ensure_future
-from inspect import isawaitable
+from inspect import isasyncgen, isawaitable
 from typing import Any, AsyncGenerator, Awaitable, List, Optional, Sequence, Type, cast
 
 from graphql import (
     DocumentNode,
     ExecutionContext,
-    ExecutionResult,
     GraphQLError,
     GraphQLSchema,
     TypeInfo,
-    execute,
     parse,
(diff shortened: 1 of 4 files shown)


ariadne/graphql.py
```diff
@@ -1,22 +1,27 @@
...
     GraphQLError,
     GraphQLSchema,
     TypeInfo,
-    execute,
     parse,
-    subscribe as _subscribe,
 )
+from graphql import subscribe as _subscribe
 from graphql.execution import MiddlewareManager
+from graphql.execution import execute as execute_sync
```
rafalp:
```suggestion
from graphql.execution import MiddlewareManager, execute as execute_sync
```

rafalp:
Looking good so far. We'll also need tests suite for this and entry in `CHANGELOG.md`

squidsoup:
Is this still being worked on? @defer is a crucial pattern for implementing the graphql 1 view, 1 query pattern, particularly when that query has expensive fields.

rafalp:
This work is currently on hold, but PR that would continue this work to completion (eg. rebase on master, optimize, add tests and error handling) is welcome.

rafalp:
There's work happening on adding `@defer` to the `graphql-core` itself.

levrik:
@rafalp It seems like `@defer` support is available in `graphql-core` nowadays. Is it working with Ariadne already?

PR #280, opened 2019-12-18
Allow the default_literal_parser to accept arguments to be passed into it.

nanorepublica:
I came across this when trying to add filters to one of the fields in my schema.

eg.
```
query($userId: ID!) {
  user(id: $userId) {
    game(
      filter: {
        start: {
          gt: "2019-10-28T11:29:39.803866-07:53"
          lt: "2019-10-29T11:29:39.803866-07:53"
        }
      }
    ) {
      start
      end
    }
  }
}
```

This was throwing the error message: 
`graphql.error.graphql_error.GraphQLError: Argument 'filter' has invalid value {start: {gt: "2019-10-29T11:29:39.803866-07:53"}}.`

Digging into the graphql library code (chucking in some print statements) revealed the default_literal_parser to be the cause with not enough arguments being specified.

I suspect that the extra arguments ought to be passed on to the `value_parser` function but I am not familiar with the internals of Ariadne yet!

Let me know what else needs to be added (I assume at least a regression test?)

### DIFF
--- a/ariadne/scalars.py
+++ b/ariadne/scalars.py
@@ -86,7 +86,7 @@ SCALAR_AST_NODES = (BooleanValueNode, FloatValueNode, IntValueNode, StringValueN
 def create_default_literal_parser(
     value_parser: GraphQLScalarValueParser,
 ) -> GraphQLScalarLiteralParser:
-    def default_literal_parser(ast, _arguments):
+    def default_literal_parser(ast, _arguments=None):
         return value_parser(ast.value)
 
     return default_literal_parser
--- a/ariadne/scalars.py
+++ b/ariadne/scalars.py
@@ -86,7 +86,8 @@ SCALAR_AST_NODES = (BooleanValueNode, FloatValueNode, IntValueNode, StringValueN
 def create_default_literal_parser(
     value_parser: GraphQLScalarValueParser,
 ) -> GraphQLScalarLiteralParser:
-    def default_literal_parser(ast):
+    def default_literal_parser(ast, *args, **kwargs):
+        print(args, kwargs)
         return value_parser(ast.value)
 
     return default_literal_parser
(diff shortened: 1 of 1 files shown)


rafalp:
Hey!

Can you please share your schema and describe what you are trying to achieve with it?

nanorepublica:
Hi

I have a fairly large schema so here is the relevant portion that is directly related to the query above:
```
input DateTimeQueryFilter {
    eq: DateTime
    gt: DateTime
    gte: DateTime
    lt: DateTime
    lte: DateTime
}

input GameFilter {
    start: DateTimeQueryFilter
    end: DateTimeQueryFilter
}

type User implements Node {
    id: ID!
    name: String!
    epics(filter: GameFilter): [Game]
}

type Game implements Node {
	id: ID!
	start: DateTime!
	end: DateTime
}
```

To give a bit of context, a `User` starts (`start`) and finishes (`end`) many `Games`. I am providing a flexible ability to filter the games returned by either field. eg Return games started after this date or Games started between X and Y etc...

I have taken inspiration for this schema model from GastbyJS, in how they handle filtering.

The DateTime scalar is the one provided in the Django contrib module.

tkamenoko:
hello, I faced the same issue and tried to find the cause.

https://github.com/graphql-python/graphql-core-next/blob/8653ef20f196c55b60522cacfc6691c202857a5e/src/graphql/utilities/value_from_ast_untyped.py#L10-L12

Looking at that code, `GraphQLScalarLiteralParser` needs 2 positional arguments, but [`ariadne` document](https://ariadnegraphql.org/docs/0.6.0/scalars#scalars-as-input) tells like this:

>literal_parser(ast) that will be used when the scalar value is passed as part of query content (e.g. { stories(publishedOn: "2018-10-26T17:45:08.805278") { ... } }).

IMO, we should fix the wrong point of the document.

nanorepublica:
I will be happy to update the relevant documentation, but only if this PR gets merged given the documentation resides in a separate repository.

rafalp:
Thanks for bringing this issue to our attention!

Sadly its holiday season at the moment so this will likely have to wait to 2020 before anyone will be able to take a look at this. :(

In meantime if you are blocked by this please use custom `ScalardType`in lieu of Ariadne.

rafalp:
Thank you for PR. I've did research on when `parse_literal` is used and found that we could fix this by _removing_ code from Ariadne.

nanorepublica:
Amazing! Always good when you can remove code!

PR #285, opened 2019-12-27
Apollo federation support

delyanr:
All,

Please find an implementation/port of the federation specifications based on [apollo-federation](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/). The code exposes `make_federated_schema` to build the federated schema, `FederatedObjectType` and `FederatedInterfaceType` to allow the definition of reference resolvers on graphql objects and interfaces respectively (see the specs for more details).

The above allows multiple (ariadne) graphql servers, e.g. in microservice architecture, to be federated by a gateway that funnels queries to the underlying servers, instead of the old (and clumsy) practice of schema stitching.

Included are also some basic tests (mostly based on the apollo team's own examples and tests). Finally, I'm not sure how to update the documentation, so please let me know if that's something I can help with (although the above link provides good enough information).

Regards!


delyanr:
I don't get any pylint issues locally - seems like there are more rules under py3.8; is this correct?

rafalp:
Thank you for contributing!

@salwator I know you've been working on GraphQL federation recently. Could you see if this work can be used to replace apollo-server in Federation?

delyanr:
Just fixed a crash in situation where a reference resolver returns None. This is good to go in my head.

bogdal:
@delyanr I think we can just disable the `too-many-ancestors` checker. Could you add `R0901` to the [Pylint config file](https://github.com/mirumee/ariadne/blob/master/.pylintrc#L6)?

delyanr:
@bogdal Done.

PR #304, opened 2020-01-28
Add Flask Blueprint

nanorepublica:
Adding a Flask Blueprint to give an easier quickstart for anyone wanting to use Flask as their server of choice.

This also includes file upload functionality.

## Todo

* [ ] Add tests
* [ ] Update documentation

### DIFF
--- a/ariadne/contrib/flask/blueprints.py
+++ b/ariadne/contrib/flask/blueprints.py
@@ -1,0 +1,39 @@
+import json
+from ariadne import graphql_sync, combine_multipart_data
+from ariadne.constants import PLAYGROUND_HTML
+from flask import request, jsonify, current_app, Blueprint
+
+
+graphql = Blueprint("graphql", __name__, url_prefix="/graphql")
+
+
+@graphql.route("/", methods=["GET"])
+def graphql_playground():
+    # On GET request serve GraphQL Playground
+    # You don't need to provide Playground if you don't want to
+    # but keep on mind this will not prohibit clients from
+    # exploring your API using desktop GraphQL Playground app.
+    return PLAYGROUND_HTML, 200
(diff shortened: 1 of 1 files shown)


rafalp:
Thanks for taking time to contribute. 👍

We've been discussing our approach to 3rd party web frameworks few times and decided that we will want to limit `ariadne.contrib` to built-in extensions and utilities like ApolloTracing, GraphQL Federations or query complexity validator, while integrations with 3rd party frameworks (such as Django or Flask) should live as separate repos (eg. `ariadne-flask`).

The reasoning for this is that the maintenance burden has become little too much than we want to, and it prevents us from delivering new features as fast as we would want to. For example, when a new option is implemented to query executor, we now need to update `ASGI`, `WSGI`, `contrib.django` (and potentially `contrib.flask`). Our tests for those already make for a considerable part of library's tests suite. The story is very similar for our docs.

I am also worried that the existence of `ariadne.contrib` is sending the wrong message to the community - I've seen people say "I've wanted do XYZ myself but I've didn't do anything in case Ariadne will add it to contrib one day". This attitude is hampering growth :(

Thank you for your understanding!

nanorepublica:
Completely understand! This was currently just copy-paste from an internal repo I am developing on right now.

An idea that might be a good halfway house (although I am sure you might have already thought and discussed it...) is to move Ariadne and it's docs to a separate organisation which would make finding/maintaining the related projects easier? 

Let me know what you think otherwise I am happy to start a new repo with this code in it.

rafalp:
Just to follow up on this, I am perfectly happy with PR to Ariadne docs that updates flask's page to point users to your package and shows example how to add GraphQL API to flask application using your project.

PR #315, opened 2020-02-07
Initial groundwork for editable / disabling graphql playground

thebenwaters:
Please see initial work for making the playground editable / ability to disable

1) I didn't think about supporting python3.6 (could add variable dataclass backport for 3.6)
2) I don't really know how to use pytest / snapshots and would love some pointers, but I'll try and get tests in
3) Is there a formatting / linting guide?

I initially wanted to try and implement the handling of different `content-type` headers, but I'll move that into a seperate PR.

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 from inspect import isawaitable
+from enum import Enum, auto
 from typing import (
     Any,
     AsyncGenerator,
@@ -21,7 +22,9 @@ from starlette.responses import HTMLResponse, JSONResponse, PlainTextResponse, R
 from starlette.types import Receive, Scope, Send
 from starlette.websockets import WebSocket, WebSocketState, WebSocketDisconnect
 
-from .constants import DATA_TYPE_JSON, DATA_TYPE_MULTIPART, PLAYGROUND_HTML
+from .constants import DATA_TYPE_JSON, DATA_TYPE_MULTIPART, PLAYGROUND_HTML, CONTENT_TYPE_JSON, CONTENT_TYPE_GRAPHQL, \
+    CONTENT_TYPE_TEXT_HTML, CONTENT_TYPE_TEXT_PLAIN
+from .playground import generate_playground_options, PlaygroundSettings
 from .exceptions import HttpBadRequestError, HttpError
 from .file_uploads import combine_multipart_data
 from .format_error import format_error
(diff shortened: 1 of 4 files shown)


patrys:
I think Playground should be available unless introspection is disabled. As long as introspection is enabled, any Playground or GraphiQL instance can be used to introspect and query the API.

rafalp:
Agreed with @patrys. People will expect security gains from disabling playground, when there are none. This will be a big security gotcha.

Quick look at spec shows that to disable introspection will need to assign custom resolver for `Query.__schema` returning empty list, and `Query.__type` that returns `None`. We could have `disable_introspection` option on `make_executable_schema` and do that here.

I've also noticed that commits are using weird indentation. Don't know if thats IDE or different OS, but it adds some noise to the diffs. ;)

thebenwaters:
@patrys @rafalp do y'all prefer using a middleware, or fully writing out the introspection SDL and writing the customised resolvers?

thebenwaters:
Just some examples from other projects:

https://github.com/graphql-python/graphene/issues/524
https://github.com/apollographql/apollo-server/issues/322

From core

https://github.com/graphql-python/graphql-core-next/blob/master/src/graphql/type/introspection.py

From spec

https://github.com/graphql/graphql-spec/blob/master/spec/Section%204%20--%20Introspection.md

patrys:
I remember some projects using a validator to disallow introspection. I think it's ok to reject the entire query if it makes the solution simpler to implement and maintain. I don't think introspection queries are likely to be mixed with production queries.

rafalp:
If we want to take validator route, we'll have to merge https://github.com/mirumee/ariadne/pull/185 first. That PR has lingered for a while, mostly because I wanted to add more mypy types and tests for extra paths, but I think we could merge it now and then fix/test `validators` option as new issues are found around it. I can aways follow-up with types and cleanups to query cost validator later.

thebenwaters:
If you guys are ok with it, i'd really like to go the middleware route

rafalp:
Can you say what are advantages of middleware route over query validator or simply swapping resolvers for introspection fields?

thebenwaters:
After finally going through the PR linked, I don't really see any advantages other than that just other projects implement using middleware, but validators make more sense.

thebenwaters:
So a couple options,

We can make this PR just about the "options" passed to the playground, and not disable it and wait until the validators PR is merged and open another


Second option, wait for validators pr and put it here

Lionesk:
My team is using Ariadne internally and we need to disable introspection in order to start pushing our app to production eventually. Given that this PR hasn't been touched in two months I'm working on a PR right now to disable it, going the validator route. I'm open to merging my work with what was done here, so that developers can selectively choose what to expose.

rafalp:
Superseded by #358

Thank you for putting this in the motion!

PR #336, opened 2020-03-16
Adds snake_caseing to complex values inside lists

Sonictherocketman:
The default decorator doesn't support converting values found inside lists to snake case. This leads to some weird inconsistencies when dealing with lists of custom object types. This change ensures that the snake-casing is applied to all levels of the object tree.

Example
--------

```
input Organization {
    orgName
}

input Project {
    parentOrgs: [Organization]
}

@mutation.field('createProject')
@convert_kwargs_to_snake_case
def create_project(_, info, project):
    print(project['parent_orgs'][0]['org_name'])

```

### DIFF
--- a/ariadne/utils.py
+++ b/ariadne/utils.py
@@ -33,6 +33,8 @@ def convert_kwargs_to_snake_case(func: Callable) -> Callable:
         for k, v in d.items():
             if isinstance(v, dict):
                 v = convert_to_snake_case(v)
+            if isinstance(v, list):
+                v = [convert_to_snake_case(i) for i in v]
             converted[convert_camel_case_to_snake(k)] = v
         return converted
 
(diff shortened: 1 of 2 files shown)


bogdal:
@Sonictherocketman Great job 👍  Can you add a test covering this case?

Sonictherocketman:
@bogdal Done and done. It looks like you have some tests failing (at least on my machine). Not sure if this is expected. My new case passes, but I saw a few that didn't.

Thanks for your great work!

rafalp:
Nice catch here, thank you for opening PR with fix!

Sonictherocketman:
It looks like there's a bug with this. Fix coming. Plz hold on merging.

Sonictherocketman:
My bad. Apologies on introducing that bug. I added a new test case explicitly for it.

paul121:
Woohoo! I was just going to submit an issue for this. Do we know when this will be released?

rafalp:
Ariadne 0.11 will be released soon™

...we just need to find time in all this crazyness currently going around to update changelog and write release notes 😅

PR #340, opened 2020-03-21
Roll-back #288: remove query/mutation execution over the websocket

### LINKED ISSUE #320: Subscriptions "Complete Example" breaks under 0.10.0
I'm working with Subscriptions in ariadne using the exact code from the [Complete Example](https://ariadnegraphql.org/docs/subscriptions#complete-example) in the docs. When I load the playground and issue the query `subscription { counter }`, instead of a working counter, an "unsupported operand" error is raised:

```
Traceback (most recent call last):",
  File \"/Users/...REDACTED.../virtualenvs/graphql-Pfr5HTvn/lib/python3.7/site-packages/graphql/execution/execute.py\", line 625, in resolve_field_value_or_error",
    result = resolve_fn(source, info, **args)",
  File \"/Users/...REDACTED.../app/__init__.py\", line 26, in counter_resolver",
    return count + 1",
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'"
```

What's wild is that the example in the docs works with ariadne 0.8.0 and 0.9.0, but not 0.10.0. I'm not familiar enough with the internals, though, to spot the issue in the [0.9.0...0.10.0 diff](https://github.com/mirumee/ariadne/compare/0.9.0...0.10.0).

(For completeness, I'm using Python 3.7.6, ariadne 0.10.0, under uvicorn 0.11.2 and gunicorn 20.0.4.)

### LINKED ISSUE #323: Rollback #288: support for GraphQL queries and mutations over websocket connections.
In #288 we've added support for executing regular GraphQL queries/mutations over websocket connection. This change, however, did break the subscriptions for some folk, doesn't work with our docs, and requires specifying operation name to run subscription.

We'll need proper fix for the issue (eg. moving query parser above the level from `graphql()` as proposed in #320), but in meantime we should rollback this change so it's not confusing to folk anymore.

rafalp:
This PR fixes #323 by rolling back changes from #288.

Also fixes #320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 0.11.0 (UNRELEASED)
+
+- Removed support for sending queries and mutations via WebSocket.
+
+
 ## 0.10.0 (2020-02-11)
 
 - Added support for `Apollo Federation`.
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -25,9 +25,9 @@ from .constants import DATA_TYPE_JSON, DATA_TYPE_MULTIPART, PLAYGROUND_HTML
 from .exceptions import HttpBadRequestError, HttpError
 from .file_uploads import combine_multipart_data
 from .format_error import format_error
-from .graphql import graphql, subscribe, validate_data
+from .graphql import graphql, subscribe
 from .logger import log_error
-from .types import ContextValue, ErrorFormatter, Extension, GraphQLResult, RootValue
+from .types import ContextValue, ErrorFormatter, Extension, RootValue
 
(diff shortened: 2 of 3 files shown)


NyanKiyoshi:
Closes https://github.com/mirumee/ariadne/issues/320 too?

rafalp:
@NyanKiyoshi indeed it does, but it was such an edge case it didn't come to mind that we actually should have a test for this.

rafalp:
Added one more test to websocket handling in ASGI, so now we have three:

- when client's payload is `{"query": "subscription { field }"}`
- when client's payload is `{"query": "subscription { field }", "operationName": null}`
- when client's payload is `{"query": "subscription OperationName { field }", "operationName": "OperationName"}`

Each of those is valid payload sent by GraphQL clients.

PR #357, opened 2020-04-15
Use graphql.utilities.value_from_ast_untyped to parse ast in default literal parser

### LINKED ISSUE #356: Exception in default_literal_parser() when ValueNode has no "value" member
The [default_literal_parser() function](https://github.com/mirumee/ariadne/blob/master/ariadne/scalars.py#L90) in ariadne.scalars expects nodes to have a "value" member.  However, this is not the case with, for example, `ObjectValueNode` or `ListValueNode`.  This causes an exception when trying to pass such nodes.

My suggestion is to use `graphql.utilities.value_from_ast_untyped` instead:
```
return value_parser(value_from_ast_untyped(ast))
``` 

I'm happy to do a PR if you guys like this change.

markedwards:
The `default_literal_parser()` function in `ariadne.scalars` returns `ast.value`. However, in the case that `ast` is a node such as `ObjectValueNode` or `ListValueNode`, this causes an exception because these nodes do not have a value member.

Fixed by parsing `ast` with `graphql.utilities.value_from_ast_untyped`.

Fixes #356

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Added `validation_rules` option to query executors as well as ASGI and WSGI apps and Django view that allow developers to include custom query validation logic in their APIs.
 - Added `validation.cost_validator` query validator that allows developers to limit maximum allowed query cost/complexity.
+- `ScalarType` default literal parser now uses `graphql.utilities.value_from_ast_untyped` to parse the AST.
 
 
 ## 0.11.0 (2020-04-01)
--- a/ariadne/scalars.py
+++ b/ariadne/scalars.py
@@ -14,6 +14,7 @@ from graphql.type import (
     GraphQLScalarValueParser,
     GraphQLSchema,
 )
+from graphql.utilities import value_from_ast_untyped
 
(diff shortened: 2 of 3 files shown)


markedwards:
@rafalp let me know if I need to do anything else to wrap this up.  Thanks!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Added `validation_rules` option to query executors as well as ASGI and WSGI apps and Django view that allow developers to include custom query validation logic in their APIs.
 - Added `validation.cost_validator` query validator that allows developers to limit maximum allowed query cost/complexity.
+- `ScalarType` default literal parser now uses `graphql.utilities.value_from_ast_untyped` to parse the ast.
```
rafalp, line 7:
```suggestion
- `ScalarType` default literal parser now uses `graphql.utilities.value_from_ast_untyped` to parse the AST.
```

markedwards: @rafalp change made

rafalp:
Thank you!

PR #358, opened 2020-04-15
Add feature to disable introspection

Lionesk:
Here is my attempt at implementing the introspection disabling feature. I opted to go with a validator approach, as discussed in https://github.com/mirumee/ariadne/pull/315. When introspection is disabled, any operation that includes introspection fields will fail.

Some insight into some of the decisions I made:
1. Since validation rules passed into either the wsgi or asgi apps are optionally callables, I opted to pass down the responsibility of adding the `IntrospectionDisabledRule` in `graphql` and `graphql_sync` once we can be sure they've been unpacked.
2. Added a more explicit check for whether something is an _introspection key_ based on definitions within `graphql_core`.
3. Disabling the playground when the introspection flag is set to false. The playground will not be functional without the ability to introspect the schema.

I added some very straightforward tests, in line with the others I was able to find. Let me know if you would like some more in depth testing.


tests/tracing/test_utils.py
```diff
@@ -15,19 +15,19 @@ def test_util_formats_info_path_value_into_reversed_list():
 
 
 def test_introspection_check_returns_true_for_introspection_field():
-    path = Mock(key="__types")
+    path = Mock(key="__type")
```
Lionesk, line 18: `__types` is not an introspection field, as per the specification. This came up given that I made the checking more specific.

Lionesk:
I just realized I didn't cover the websocket workflow, so will be reorienting for that, and given this, I will be moving the addition of the `IntrospectionDisabledRule` to the validation rules up one level, to avoid too much code duplication.

tests/tracing/test_utils.py
rafalp: Nice catch!

ariadne/asgi.py
```diff
@@ -121,7 +123,9 @@ async def get_middleware_for_request(
 
     async def handle_http(self, scope: Scope, receive: Receive, send: Send):
         request = Request(scope=scope, receive=receive)
-        if request.method == "GET":
+        if (
+            request.method == "GET" and self.introspection
+        ):  # only render playground when introspection is enabled
```
rafalp, line 128:
```suggestion
        if request.method == "GET" and self.introspection:
            # only render playground when introspection is enabled
```

ariadne/validation/introspection.py
```diff
@@ -0,0 +1,19 @@
+from graphql import GraphQLError
```
rafalp, line 1: Could you rename the module to `instrospection_disabled`?

tests/test_graphql.py
```diff
@@ -26,6 +26,17 @@ def test_graphql_sync_uses_validation_rules(schema):
     assert result["errors"][0]["message"] == "Invalid"
 
 
+def test_graphql_sync_prevents_introspection(schema):
```
rafalp, line 29:
```suggestion
def test_graphql_sync_prevents_introspection_query_when_option_is_disabled(schema):
```

tests/test_graphql.py
```diff
@@ -42,6 +53,18 @@ async def test_graphql_uses_validation_rules(schema):
     assert result["errors"][0]["message"] == "Invalid"
 
 
+@pytest.mark.asyncio
+async def test_graphql_prevents_introspection(schema):
```
rafalp, line 57:
```suggestion
async def test_graphql_prevents_introspection_query_when_option_is_disabled(schema):
```

tests/test_graphql.py
```diff
@@ -57,3 +80,15 @@ async def test_subscription_uses_validation_rules(schema):
     )
     assert not success
     assert result[0]["message"] == "Invalid"
+
+
+@pytest.mark.asyncio
+async def test_subscription_prevents_introspection(schema):
```
rafalp, line 86:
```suggestion
async def test_subscription_prevents_introspection_query_when_option_is_disabled(schema):
```

rafalp:
Looking good so far, thank you! 

Do you think you'll be able to also add `instrospection` option to Django view in `ariadne.contrib.django`?

Lionesk:
@rafalp would you like me to open a corresponding PR on https://github.com/mirumee/ariadne-website to update the docs?

rafalp:
@Lionesk that would be perfect!

ariadne/graphql.py
```diff
@@ -17,6 +17,7 @@
 from graphql.validation import specified_rules, validate
 from graphql.validation.rules import RuleType
 
+from .validation.introspection_disabled import IntrospectionDisabledRule
```
rafalp, line 20: Please move this import under `from .types`.

ariadne/graphql.py
```diff
@@ -136,6 +145,12 @@ def graphql_sync(
 
             if callable(validation_rules):
                 validation_rules = validation_rules(context_value, document, data)
+            if not introspection:
+                validation_rules = (
+                    list(validation_rules) + [IntrospectionDisabledRule]
+                    if validation_rules
+                    else [IntrospectionDisabledRule]
+                )
```
rafalp, line 153: Looks too boiler-platey being repeated three times in file, and having `[IntrospectionDisabledRule]` six times total.

```suggestion
            if not introspection:
                validation_rules = disable_query_introspection(validation_rules)


def disable_query_introspection(validation_rules: Optional[ValidationRules]) -> ValidationRules:
    if validation_rules:
        return validation_rules = [IntrospectionDisabledRule]
    return [IntrospectionDisabledRule]
```

ariadne/validation/introspection_disabled.py
```diff
@@ -0,0 +1,19 @@
...
+
+class IntrospectionDisabledRule(ValidationRule):
+    def enter_field(self, node: FieldNode, *_args):
+        field_name = node.name.value
+        if not is_introspection_key(field_name):
+            return
+
+        self.report_error(
+            GraphQLError(
+                f"Introspection has been disabled, and {field_name} is an introspection field",
```
rafalp, line 16: How about
```suggestion
                f"Cannot query '{field_name}': introspection is not available.",
```

ariadne/graphql.py
```diff
@@ -58,9 +60,22 @@ async def graphql(
             document = parse_query(query)
 
             if callable(validation_rules):
-                validation_rules = validation_rules(context_value, document, data)
+                # validation_rules are optionally callable, so evaluating them here changes their
+                # type signature
```
rafalp, line 64: Isn't this issue solved by `cast`?

```
validation_rules = cast(
    Optional[Sequence[RuleType]],
    validation_rules(context_value, document, data),
)
```

Lionesk: It is! Thank you.

ariadne/graphql.py
```diff
@@ -343,8 +363,9 @@ def validate_operation_name(operation_name) -> None:
         raise GraphQLError('"%s" is not a valid operation name.' % operation_name)
 
 
-def validate_context_value(context_value) -> None:
```
Lionesk, line 346: I noticed this was not being used.

rafalp:
I've just realized: if we are doing `validate_query`, why not just update it's signature? That would minimalize amount of boilerplate added to `graphql_` methods and let us drop `disable_query_introspection `:

```
def validate_query(
    schema: GraphQLSchema,
    document_ast: DocumentNode,
    rules: Optional[Sequence[RuleType]] = None,
    type_info: Optional[TypeInfo] = None,
    enable_introspection: bool = True,
) -> List[GraphQLError]:
```

I've also noticed that we are calling `validate()` instead of our `validate_query` inside `subscription()`. I'm wondering if that's omission, or some gotcha in subscription handling? 🤔

Sorry for being so late to spot this, but I feel we are step away from this! Thank you for the effort you are putting into this!

Lionesk:
@rafalp Followed your suggestion, and seems that it was indeed an omission in the subscription flow, as none of the tests broke.

rafalp:
Excellent work, thank you!

PR #392, opened 2020-06-23
Django error formatter

maxmorlocke:
Ariadne's documentation is rightfully opinionated on the communication of exceptions from the server and strongly suggests protecting the errors key.  It rightfully suggests replacing the default error formatter.  To date, no preexisting sanely configured, easily overridable solution exists for communicating messages to users in the event of validation errors or other common problematic workflows (e.g. normal 401/403 type events).  

In this PR, I have provided a simple solution that includes short, descriptive error messaging tied to HTTP-style messages for errors.  I have also provided a path for user facing error messaging to be communicated.  Communicating client-facing errors from server to client is a normal expectation, and providing a field exclusively for that allows for client-side developers to easily distinguish between GraphQL errors that are *NOT* user facing and ValidationErrors, etc. that most assuredly are.  

Fair credit to the following post for providing some inspiration:

https://medium.com/@tarkus/validation-and-user-errors-in-graphql-mutations-39ca79cd00bf

The major difference is I use "non_field_errors" instead of "", which both makes intent more clear and follows standard django patterns.

Documentation PR: https://github.com/mirumee/ariadne-website/pull/58

### DIFF
--- a/ariadne/contrib/django/constants.py
+++ b/ariadne/contrib/django/constants.py
@@ -1,0 +1,19 @@
+FORMATTED_ERROR_MESSAGES = {
+    "ValidationError": {"short": "Invalid input", "details": None,},
+    "ObjectDoesNotExist": {
+        "short": "Not found",
+        "details": "We were unable to locate the resource you requested.",
+    },
+    "NotAuthenticated": {
+        "short": "Unauthorized",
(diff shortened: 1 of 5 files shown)


kedlav:
This should be ready for review

PR #393, opened 2020-06-24
graphql-core 3.1

### LINKED ISSUE #345: Ariadne is not compatible with graphql-core 3.1
Hi folks,

I'm hitting this error with the very example at https://www.starlette.io/

```python
from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.routing import Route


async def homepage(request):
    return JSONResponse({'hello': 'world'})


app = Starlette(debug=True, routes=[
    Route('/', homepage),
])
```

```
  File "./example.py", line 1, in <module>
    from ariadne import QueryType, make_executable_schema
  File "x/lib/python3.8/site-packages/ariadne/__init__.py", line 11, in <module>
    from .graphql import graphql, graphql_sync, subscribe
  File "x/lib/python3.8/site-packages/ariadne/graphql.py", line 18, in <module>
    from graphql.validation.rules import RuleType
```

```
❯ python --version
Python 3.8.2

❯ pip freeze | grep "ariadne"
ariadne==0.10.0

❯ python -c "import graphql; print(graphql.__version__)"
3.1.0b2
```

Digging a bit:

Looks like this [commit](https://github.com/graphql-python/graphql-core/commit/2744051ec8be7dd1d2eae6d08947106f6c12e86a#diff-4c0d14e6bf0dcc681bb5ad2994e432e3L13) from `graphql-python` removed `RuleType` from `graphql.validation.rules`

Thanks!

nikordaris:
Fixes #345


ariadne/executable_schema.py
```diff
@@ -48,16 +46,6 @@ def join_type_defs(type_defs: List[str]) -> str:
...
-    schema = build_ast_schema(ast)
-    extension_ast = extract_extensions(ast)
-
-    if extension_ast.definitions:
-        schema = extend_schema(schema, extension_ast)
-
-    return schema
-
-
 EXTENSION_KINDS = [
```
connorbrinton, line 49: @nikordaris I think this list of `EXTENSION_KINDS` can also be removed, from what I can tell it's only used in `extract_extensions`, which is no longer present.

Hope this gets merged and released soon! 😄

nikordaris: Nice catch, I'll take a look tomorrow morning and get it removed.

delyanr:
Any plans to have this merged? Thanks

rafalp:
Hi @nikordaris 

Thanks for opening this PR, its very helpful! Do you think you'll find time to address @connorbrinton comment? Other than that this is looking good to me :)

Thanks again!

nikordaris:
@rafalp yup, just removed that unused extension kinds list.

rafalp:
Looks like test is failing on snapshot not matching. Can you rebase on master and update the snapshot?

nikordaris:
so i'm not getting any errors locally about snapshot. in python 3.8 i did error on missing `python-multipart` dependency so i added that to requirements-dev and the tests passed locally. we'll see what happens in travis.

https://www.starlette.io/#dependencies

nikordaris:
trying python 3.8.0 since i think that is what travis is using

rafalp:
Travis is actually running tests against three versions: 3.6, 3.7 and 3.8

setup.py
```diff
@@ -31,7 +31,7 @@
     packages=["ariadne"],
     include_package_data=True,
     install_requires=[
-        "graphql-core>=3.0.0,<3.1.0",
+        "graphql-core>=3.0.0",
```
rafalp, line 34:
```suggestion
        "graphql-core>=3.1.0",
```

rafalp (changes requested):

rafalp:
@nikordaris Travis is installing deps via setup.py which specifies 3.0.0 as requirement ;)

setup.py
nikordaris: do we want to force 3.1 on people though?

nikordaris: ah, shit. its breaking change. nvm. ya we do

rafalp:
@nikordaris one more thing about introspection failing: GraphQL-Core 3.1.2 added the `SpecifiedBy` directive, so if snapshot was updated for 3.1.0 or 3.1.1 it will not work with 3.1.2. :)

rafalp:
On second thought, this part of tests is fragile while not adding anything of value. We should remove this snapshot and use simple assert to verify that introspection query returned any `data`.

nikordaris:
(╯°□°）╯︵ ┻━┻

lol, ok i'll take a look tomorrow

nikordaris:
@rafalp ok, tests are fixed now

rafalp (approved):

rafalp:
Excellent, thank you!

PR #394, opened 2020-06-26
Django serializer mutation resolver

maxmorlocke:
This PR provides a mutation resolver that is tightly coupled with Django REST Framework's ModelSerializer.  This allows mutations to be quickly and easily configured to perform validation and create, update, and delete operations leveraging a known, highly reliable, well documented, and performant solution.

This solution follows the outline setup by DRF's ModelViewSet.

Documentation PR: https://github.com/mirumee/ariadne-website/pull/59

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,6 +1,6 @@
 [MASTER]
 ignore=snapshots
-load-plugins=pylint.extensions.bad_builtin, pylint.extensions.mccabe
+load-plugins=pylint.extensions.bad_builtin, pylint.extensions.mccabe, pylint_django
 
 [MESSAGES CONTROL]
 disable=C0103, C0111, C0330, C0412, C1001, E1004, I0011, R0101, R0201, R0801, R0901, R0902, R0903, R0912, R0913, R0914, R0915, R1260, W0231, W0232, W0621, W0703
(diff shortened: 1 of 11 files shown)


kedlav:
This should be ready for review

rafalp:
We'll pass on this landing in `ariadne.contrib.django`, with the caveat being that we want to split all django-specific code from Ariadne into separate package (like `django-ariadne`). Merging this only to remove it in next release seems counterproductive.

I'll see when I can get the ball rolling on this. Likely will share more details on spectrum when I have the details.

kedlav:
> We'll pass on this landing in `ariadne.contrib.django`, with the caveat being that we want to split all django-specific code from Ariadne into separate package (like `django-ariadne`). Merging this only to remove it in next release seems counterproductive.
> 
> I'll see when I can get the ball rolling on this. Likely will share more details on spectrum when I have the details.

That makes sense.  Let me know if there's anything I can do to help.  I'm happy to be a contributor in whatever way y'all feel is best to package things up.  We're really happy as ariadne+django users.

nanorepublica:
I would be happy to be part of this conversation as I have built a similar MutationResolver that couples with DRF. I would need to check if I have permission to share the code first but definitely want to help make ariadne + django better!

PR #406, opened 2020-08-03
Expose websockets for auth (improved)

Gnork:
I have improved the open pull request https://github.com/mirumee/ariadne/pull/355.

1. I rebased the original pull request to the current master branch
2. My code can handle async or normal `on_connect` and `on_close` functions as requested in https://github.com/mirumee/ariadne/pull/355#discussion_r416528840
3. The `context_value` is now created earlier in the `websocket_server` and is passed to all the other functions. It is also passed  to `on_connect` and `on_close`, instead of the websocket object (because this object is included in the context by default anyway). The user can therefore add content to the context inside of `on_connect` (e.g. message payload) that can then be reused in a resolver.

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -28,11 +28,11 @@ from .format_error import format_error
 from .graphql import graphql, subscribe
 from .logger import log_error
 from .types import (
-    OnConnect,
-    OnClose,
     ContextValue,
     ErrorFormatter,
     Extension,
+    OnClose,
+    OnConnect,
     RootValue,
     ValidationRules,
 )
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -239,7 +239,9 @@ class GraphQL:
                 and websocket.application_state != WebSocketState.DISCONNECTED
             ):
                 message = await websocket.receive_json()
-                await self.handle_websocket_message(message, websocket, subscriptions, context_value)
+                await self.handle_websocket_message(
+                    message, websocket, subscriptions, context_value
+                )
         except WebSocketDisconnect:
             pass
         finally:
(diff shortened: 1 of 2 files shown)


ariadne/asgi.py
```diff
@@ -27,7 +27,15 @@
 from .format_error import format_error
 from .graphql import graphql, subscribe
 from .logger import log_error
-from .types import ContextValue, ErrorFormatter, Extension, RootValue, ValidationRules
+from .types import (
+    OnConnect,
+    OnClose,
```
rafalp, line 32: Please order imports alphabetically.

loic001:
Hi, any news about that?

ariadne/asgi.py
```diff
@@ -58,6 +66,8 @@ def __init__(
         self,
         schema: GraphQLSchema,
         *,
+        on_connect: Optional[OnConnect] = None,
+        on_close: Optional[OnClose] = None,
```
TheDelus, line 70:
```suggestion
        on_ws_connect: Optional[OnConnect] = None,
        on_ws_close: Optional[OnClose] = None,
```
The name should represent that this only belongs to the websocket connection.

TheDelus (changes requested):
Even though I request changes here, I realy wana see this feature makes it in. It would simplify the whole auth process by a lot.

ariadne/asgi.py
rafalp: Why? Are there other connection types that GraphQL is aware of that could be confusing?

rafalp:
@TheDelus depending on who you ask either this or no support for GraphQL-core 3.1.x come up as top issue with Ariadne.

The latter will be resolved with 0.13 release, meaning I can move my attention to latter for 0.14 :)

rafalp:
I've opened new PR #488 that replaces this work with different approach.

Thanks!

PR #410, opened 2020-08-23
Fix cost multiplier to accept nested variables

tkamenoko:
This PR makes a query cost validator to accept deep nested variables like `input` type values.
For example, when a multiplier name is `foo.bar.baz` ,  the multiplier value is `{"foo" : {"bar" : {"baz" : <value>}}}` .

See https://github.com/pa-bru/graphql-cost-analysis#cost-settings-arguments

note: CRLF was used in the 2 changed files, but most of `ariadne` used LF, so I fixed that newlines.

### DIFF
--- a/ariadne/validation/query_cost.py
+++ b/ariadne/validation/query_cost.py
@@ -86,9 +86,12 @@ class CostValidator(ValidationRule):
                     continue
                 field_type = get_named_type(field.type)
                 try:
-                    field_args = get_argument_values(field, child_node, self.variables)
+                    field_args: Dict[str, Any] = get_argument_values(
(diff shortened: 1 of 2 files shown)


rafalp:
Hi @tkamenoko, thank you for finding and fixing this.

Can you please rebase against master?

tkamenoko:
@rafalp Sorry for my late reply. Should I fix anything else?

rafalp:
Tests suite is failing because of mypy changes, but tests themselves are passing. I'll still approve and merge this. Thank you!

rafalp (approved):

PR #427, opened 2020-09-25
Default union type resolver

pawelrubin:
## Problem

Currently handling **UnionTypes** requires writing a boiler-plate type resolver. It tends to be irritating, especially for big unions.

Example from [docs](https://ariadnegraphql.org/docs/next/unions#union-example):
```python
from ariadne import UnionType

error = UnionType("Error")

@error.type_resolver
def resolve_error_type(obj, *_):
    if isinstance(obj, ValidationError):
        return "ValidationError"
    if isinstance(obj, AccessError):
        return "AccessError"
    return None
```

## Solution
Default type resolver. You can still set a custom `type_resolver`, but the common behaviour is the default.

```python
def default_type_resolver(obj: Any, *_: Any) -> str:
    return type(obj).__name__
```

### DIFF
--- a/ariadne/interfaces.py
+++ b/ariadne/interfaces.py
@@ -8,17 +8,20 @@ from graphql.type import (
 )
 
 from .objects import ObjectType
-from .types import Resolver
+from .resolvers import default_type_resolver
+from .types import TypeResolver
 
 
 class InterfaceType(ObjectType):
-    _resolve_type: Optional[Resolver]
+    _resolve_type: TypeResolver
 
-    def __init__(self, name: str, type_resolver: Optional[Resolver] = None) -> None:
+    def __init__(
+        self, name: str, type_resolver: TypeResolver = default_type_resolver
+    ) -> None:
         super().__init__(name)
         self._resolve_type = type_resolver
 
-    def set_type_resolver(self, type_resolver: Resolver) -> Resolver:
+    def set_type_resolver(self, type_resolver: TypeResolver) -> TypeResolver:
         self._resolve_type = type_resolver
         return type_resolver
 
--- a/ariadne/resolvers.py
+++ b/ariadne/resolvers.py
@@ -63,3 +63,7 @@ def is_default_resolver(resolver: Resolver) -> bool:
     if resolver == default_field_resolver:
         return True
     return hasattr(resolver, "_ariadne_alias_resolver")
+
+
+def default_type_resolver(obj: Any, *_: Any) -> str:
+    return type(obj).__name__
--- a/ariadne/types.py
+++ b/ariadne/types.py
@@ -43,6 +43,11 @@ ValidationRules = Union[
 ]
 
 
+class TypeResolver(Protocol):
+    def __call__(self, obj: Any, *_: Any) -> Optional[str]:
+        pass  # pragma: no cover
+
+
 class Extension(Protocol):
     def request_started(self, context: ContextValue):
         pass  # pragma: no cover
--- a/ariadne/unions.py
+++ b/ariadne/unions.py
@@ -2,17 +2,17 @@ from typing import Optional, cast
 
 from graphql.type import GraphQLNamedType, GraphQLUnionType, GraphQLSchema
 
-from .types import Resolver, SchemaBindable
+from .types import SchemaBindable, TypeResolver
 
 
 class UnionType(SchemaBindable):
-    _resolve_type: Optional[Resolver]
+    _resolve_type: Optional[TypeResolver]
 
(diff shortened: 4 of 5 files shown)


pawelrubin:
PR ready for review, although I'm not sure why the codecov/project check failed.

rafalp:
Thanks for opening this PR!

Looking good, but I'm wondering if we could (or should) somehow combine this with default behavior provided by the GraphQL executor when no type resolver is defined: accessing object's `__typename` attribute (or dict key) to see if typename is already defined on object.

pawelrubin:
Hi @rafalp. Thanks for the quick response.

I'm not sure if I follow. Do you mean to check for the presence of `obj["__typename"]` in the `make_default_type_resolver`?

rafalp:
Yes.

pawelrubin:
That's an interesting point. I'll look into that.

patrys:
I think it may actually be easier to explicitly define the `__typename` on these types instead of writing a custom union resolver.

pawelrubin:
The [graphql-core](https://github.com/graphql-python/graphql-core/blob/master/src/graphql/execution/execute.py#L1224) codebase provides default resolvers. Both for fields and types.

```python

def default_type_resolver(
    value: Any, info: GraphQLResolveInfo, abstract_type: GraphQLAbstractType
) -> AwaitableOrValue[Optional[Union[GraphQLObjectType, str]]]:
    """Default type resolver function.

    If a resolve_type function is not given, then a default resolve behavior is used
    which attempts two strategies:

    First, See if the provided value has a ``__typename`` field defined, if so, use that
    value as name of the resolved type.

    Otherwise, test each possible type for the abstract type by calling
    :meth:`~graphql.type.GraphQLObjectType.is_type_of` for the object
    being coerced, returning the first type that matches.
    """
    # First, look for `__typename`.
    type_name = get_typename(value)
    if isinstance(type_name, str):
        return type_name

    # Otherwise, test each possible type.
    possible_types = info.schema.get_possible_types(abstract_type)
    is_awaitable = info.is_awaitable
    awaitable_is_type_of_results: List[Awaitable] = []
    append_awaitable_results = awaitable_is_type_of_results.append
    awaitable_types: List[GraphQLObjectType] = []
    append_awaitable_types = awaitable_types.append

    for type_ in possible_types:
        if type_.is_type_of:
            is_type_of_result = type_.is_type_of(value, info)

            if is_awaitable(is_type_of_result):
                append_awaitable_results(cast(Awaitable, is_type_of_result))
                append_awaitable_types(type_)
            elif is_type_of_result:
                return type_

    if awaitable_is_type_of_results:
        # noinspection PyShadowingNames
        async def get_type():
            is_type_of_results = await gather(*awaitable_is_type_of_results)
            for is_type_of_result, type_ in zip(is_type_of_results, awaitable_types):
                if is_type_of_result:
                    return type_

        return get_type()

    return None


def default_field_resolver(source: Any, info: GraphQLResolveInfo, **args: Any) -> Any:
    """Default field resolver.

    If a resolve function is not given, then a default resolve behavior is used which
    takes the property of the source object of the same name as the field and returns
    it as the result, or if it's a function, returns the result of calling that function
    while passing along args and context.

    For dictionaries, the field names are used as keys, for all other objects they are
    used as attribute names.
    """
    # Ensure source is a value for which property access is acceptable.
    field_name = info.field_name
    value = (
        source.get(field_name)
        if isinstance(source, dict)
        else getattr(source, field_name, None)
    )
    if callable(value):
        return value(info, **args)
    return value
```


I think they could be imported by **ariadne**. What do You think @rafalp @patrys?

pawelrubin:
I've added the `default_type_resolver`. Since the default behavior is the same as the one achieved with `make_default_type_resolver`, I've deleted it.

I think this is ready for the review.

pawelrubin:
Hi @rafalp @patrys. Any chance for a review/comments? 🙏

rafalp (changes requested):
Default type resolver should still respect `__typename` dict key/obj attribute.

rafalp:
Going for object's name is acceptable fallback, but only if there's no `__typename` attribute (or dict key when resolver's return value is dict).

I'm also unsure what happens when type resolver returns type name that's not present in the schema. This needs to be checked because for dict typename will resolve to `dict` (it should resolve to `None` instead).

rafalp:
As said above having `__typename` may be easiest solution for the problem. In `graphql-modules` we may eventually implement option for types do declare their python representation enabling us to write "clean" default resolver for it's `UnionType`.

Thank you for contributing and sorry for taking two years for this :(

PR #443, opened 2020-11-12
Change supported starlette versions to include 0.14

cancan101:


rafalp:
Did you mean include `0.15`? ;)

cancan101:
> Did you mean include `0.15`? ;)

I don't think so. 0.14 has been released whereas 0.15 has not: https://www.starlette.io/release-notes/

cancan101:
To avoid issues with pre-releases this would probably be better written as `<= 0.14.99`, but I kept as is for consistency.

rafalp:
Ah, right, my mistake. Tests are passing, but linters are failing because updated Mypy is finding types errors. I'll accept this still.

rafalp (approved):

cancan101:
Should this be accepted, though? It seems like the API in starlette might have changed causing these errors?

rafalp:
No - only errors linters found are in types mismatch between the library and graphql-core.

cancan101:
Oh, interesting. What caused that, a change in the deps for starlette? and either way, any reason for concern here?

rafalp:
I've merged the PR from dependabot bumping mypy to latest versions in the meantime - that caused it.

> any reason for concern here?

Nope. We have quite a good unit tests suite in place for Ariadne, and that one is sitting on the top of tests suites for graphql-core and Starlette.

PR #449, opened 2020-11-30
Ensure None entities are returned as such during federation (without appending typename)

delyanr:
Hello! My original federation PR contained an oversight on my part that wrongly added `__typename` to `None` when resolving entities. As can be seen in the [federation specs](https://www.apollographql.com/docs/federation/federation-spec/) and the original [federation-js source](https://github.com/apollographql/federation/blob/49fc528be1c2796b09fe7dff4f8cbe604b39a3fb/federation-js/src/types.ts#L44), the typename should not be added, as entities can return a non-nullable list of nullable `_Entity` types.

This PR addresses this oversight and tweaks one of the test files. Cheers!

### DIFF
--- a/ariadne/contrib/federation/utils.py
+++ b/ariadne/contrib/federation/utils.py
@@ -137,4 +137,4 @@ def add_typename_to_possible_return(obj: Any, typename: str) -> Any:
         else:
             setattr(obj, f"_{obj.__class__.__name__}__typename", typename)
         return obj
-    return {"__typename": typename}
+    return None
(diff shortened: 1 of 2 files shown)


rafalp (approved):

rafalp:
Tests are passing, but linters are failing. I'll approve this and merge it anyway after I've fixed linters.

rafalp:
I've decided to merge this first and fix linters at later date.

Thank you for contributing!

PR #451, opened 2020-12-01
Updated convert_camel_case_to_snake to handle more edge cases.

nilsdet:
Updated convert_camel_case_to_snake to not add an additional underscore, if the previous char was already one.

This is useful for example for django users, when using it's nested lookups.
The previous version of the convert_camel_case_to_snake would add an additional underscore to the string, the new one prevents this behaviour.

So passing a string like "user__accountId" now results in "user__account_id" (2 underscores) instead of "user___account_id" (3 underscores).


rafalp:
Thank you for contributing this improvement!

PR #452, opened 2020-12-03
Do not add underscores in block capitals in identifiers

BenMotz:
Do not add an underscore, if the previous character was also upper case.

This is useful if you have an acronym in an identifier. Specifically, I encountered this with the identifier `interfaceMACs`, which converted to `interface_m_a_cs`. With this change the identifier converts to `interface_macs`.

### DIFF
--- a/ariadne/utils.py
+++ b/ariadne/utils.py
@@ -11,7 +11,6 @@ def convert_camel_case_to_snake(graphql_name: str) -> str:
         if (
             i
             and c != graphql_name[i]
-            and i > 0
             and graphql_name[i - 1] != "_"
             and graphql_name[i - 1] == python_name[-1]
         ):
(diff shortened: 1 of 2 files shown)


ariadne/utils.py
```diff
@@ -8,7 +8,13 @@
 def convert_camel_case_to_snake(graphql_name: str) -> str:
     python_name = ""
     for i, c in enumerate(graphql_name.lower()):
-        if i and c != graphql_name[i] and i > 0 and graphql_name[i - 1] != "_":
+        if (
+            i
+            and c != graphql_name[i]
+            and i > 0
```
rafalp, line 14: We can drop `and i > 0` here because it doubles for initial `i` check.

rafalp (changes requested):
Thank you for stepping up to improve Ariadne!

This is looking good, just one tiny nitpick thats really on Us, but if it can be resolved in this PR too, I'll be happy to merge this!

rafalp (approved):

rafalp:
Travis tests are passing with only fails being pylint and black. I'll fix those later myself.

Thank you!

PR #465, opened 2021-01-15
Added support for numbers in camelCase variables.

nilsdet:
The previous version of the 'convert_camel_case_to_snake' utility function did not handle numbers inside the 'graphql_name'.

So passing a string like "foo365Bar" woud have resulted in "foo365_bar".

Now it would result in: "foo_365_bar"

### DIFF
--- a/tests/test_case_convertion_util.py
+++ b/tests/test_case_convertion_util.py
@@ -44,5 +44,5 @@ def test_no_underscore_added_if_previous_character_is_uppercase():
     assert convert_camel_case_to_snake("testWithUPPERPart") == "test_with_upperpart"
 
 
-def test_number_gets_treated_as_word():
+def test_digits_are_treated_as_word():
     assert convert_camel_case_to_snake("testWith365InIt") == "test_with_365_in_it"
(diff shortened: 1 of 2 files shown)


tests/test_case_convertion_util.py
```diff
@@ -42,3 +42,7 @@ def test_no_underscore_added_if_previous_character_is_an_underscore():
 
 def test_no_underscore_added_if_previous_character_is_uppercase():
     assert convert_camel_case_to_snake("testWithUPPERPart") == "test_with_upperpart"
+
+
+def test_number_gets_treated_as_word():
```
rafalp, line 47:
```suggestion
def test_digits_are_treated_as_word():
```

rafalp (approved):

rafalp:
Thats one weird conditional here, but change itself looks useful. Thank you!

ramikhalaf:
Heads up this is a breaking change when going from 0.12.0 to 0.13.0. All existing resolver arguments go from being converted to `variable1` to `variable_1`

PR #470, opened 2021-01-30
Fixed type annotation for extensions list

dkbarn:
This consolidates the definition of the ExtensionList type into ariadne.types, where it can be referenced by everything that needs it.

The type definition has been updated to support a list of both Extension classes and callables that return Extension objects. This should be allowed since all that's happening with the items in this list is calling them: https://github.com/mirumee/ariadne/blob/master/ariadne/extensions.py#L21

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -9,7 +9,6 @@ from typing import (
     List,
     Optional,
     Union,
-    Type,
     cast,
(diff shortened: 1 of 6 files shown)


rafalp (approved):

rafalp:
LGTM, thank you!

PR #482, opened 2021-03-02
Fix: 293: use python represenation of enum values for default values in schema

### LINKED ISSUE #293: `make_executable_schema`'s  omits default enum values in fields args
GraphQL enums are represented in Python via the [`GraphQLEnumType`](https://github.com/graphql-python/graphql-core-next/blob/master/src/graphql/type/definition.py#L996). If those types are created from the SDL, they define no Python representation for GraphQL enum items. Developers can use Ariadne-provided `EnumType` to specify how enum items should be represented in Python themselves, but we also include "fallback" step that fills enum values with their `str` representations.

However neither of those approaches takes care of enum items defined as default values for field arguments. Eg:

```
type Query {
    users(role: ADMIN): [User]
}
```

We will need to add additional step to `make_executable_schema` that gathers all enums from schema together with their Python representations, and then introspects schema's fields values for places where field has default value type of enum. If this is the case, we should set correct `default_value` on [`GraphQLArgument`](https://github.com/graphql-python/graphql-core-next/blob/b320331faf2fc2f4f1f6a1366f07109d1bdd44f1/src/graphql/type/definition.py#L576) in the schema.

rafalp:
Fixes #293 

This PR updates `EnumType` to scan schema for fields where Enum is used for value, and default value is defined.

**TODO**

- [x] Handle fields args `field(arg: Enum = DEFAULT)`
- [x] Handle inputs fields `arg: Enum = DEFAULT`
- [x] Handle fields inputs args `field(arg: {field: DEFAULT})`
- [ ] Nested field inputs args `field(arg: {child: {field: DEFAULT}})`
- [ ] Write tests
- [ ] Cleanup the code

### DIFF
--- a/ariadne/enums.py
+++ b/ariadne/enums.py
@@ -2,7 +2,16 @@ import enum
 
 from typing import Any, Dict, Optional, Union, cast
 
-from graphql.type import GraphQLEnumType, GraphQLNamedType, GraphQLSchema
+from graphql.type import (
+    GraphQLEnumType,
+    GraphQLInputField,
+    GraphQLInputObjectType,
+    GraphQLNamedType,
+    GraphQLObjectType,
+    GraphQLArgument,
+    GraphQLSchema,
+)
+from graphql.pyutils import Undefined
 
 from .types import SchemaBindable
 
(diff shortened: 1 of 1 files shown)


rafalp:
Superseded by #533

Issue #483, opened 2021-03-02
Add Python 3.9 to testing matrix


rafalp:
Python 3.9 has been available for quite a while and GraphQL core declares support for it since 3.1.3.

I'm expecting Ariadne to work correctly on 3.9, but it would be great if we've had tests in place to prove that.

rafalp:
Fixed on master

PR #484, opened 2021-03-04
Replace Travis-CI with GitHub actions

### LINKED ISSUE #268: Move from Travis to GitHub actions

rafalp:
Fixes #268 and #483

Removes Travis CI and adds GH actions instead.

Note: pip-compile insists on including `typed-ast` which is not supported for Python 3.8 and outright crashes installation on Python 3.9, hence the separate dev deps file for python 3.9.

Maybe we should drop Mypy/pylint/black from dev deps and install it only on CI step?


bogdal (approved):
LGTM 👍

PR #403, opened 2020-07-31
hand connection_init message payload to subscription context

Gnork:
For websocket subscriptions, GraphQL Playground sends header data (e.g. for authorization) in the `connection_init` message payload. In the current Ariadne implementation this data is discarded. My fix persists the payload from the `connection_init` message and hands it to the `start_websocket_subscription` method when the `start` message arrives. The payload data is then stored in the `context` dictionary.

I moved the method code from `handle_websocket_message` to `websocket_server` and deleted `handle_websocket_message`, because with the stateful behaviour introduced by my fix it seems more elegant this way.

PR #488, opened 2021-03-12
Add on_connect/on_disconnect hooks to ASGI app, support connection context

rafalp:
Adds `on_connect` and `on_disconnect` options to ASGI GraphQL app and introduces connection context.

# `on_connect`

`on_connect` is optional callable that is called when GraphQL client opens WebSocket connection with the server. It's called with two arguments: `WebSocket` instance and `Any` value with connection payload. It can be used to extract supported data from connection's payload to the current scope:

```python
def on_connect(ws: WebSocket, payload: Any):
    if not isinstance(payload, dict):
        return

    ws.scope["auth"] = get_user_from_token(payload.get("auth_token"))


def get_context_value(request):
    if request.scope.type == "websocket":
        return {"auth": request.scope["auth"]}
    
    return {"auth": get_user_from_token(request.headers.get("authorization")}
```

As bonus, users may raise new `WebSocketConnectionError` from `on_connect` to get nice error masse sent to client:

```python
from ariadne.asgi import WebSocketConnectionError


def on_connect(ws: WebSocket, payload: Any):
    if not isinstance(payload, dict):
        raise WebSocketConnectionError({"message": "payload object is required", "code": "payload.value_error"})

    ws.scope["auth"] = get_user_from_token(payload.get("auth_token"))
```

# `on_disconnect`

`on_disconnect` is optional callable that is called when GraphQL client closes websocket connection. Receives one argument: `WebSocket` instance.

```python
def on_disconnect(ws):
    # get some data from WS scope and maybe log it somewhere or set user from `ws.scope` to offline?
```

Supersedes #406, #403 and #355.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 0.14.0 (Unreleased)
+
+- Added `on_connect` and `on_disconnect` options to `ariadne.asgi.GraphQL`, enabling developers to run additional initialization and cleanup for websocket connections.
+
+
 ## 0.13.0 (2021-03-17)
 
 - Updated `graphQL-core` requirement to 3.1.3.
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -316,6 +316,8 @@ class GraphQL:
         self,
         websocket: WebSocket,
     ):
+        await websocket.close()
+
         try:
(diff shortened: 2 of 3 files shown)


bogdal (approved):

ammurdoch:
This is great!  When can y'all get this released?

PR #529, opened 2021-04-06
#487 Deepcopy issue on uploaded files in filter resolver

### LINKED ISSUE #487: OpenTracing plugin performs deepcopy of resolver's args, which fails when file upload for larger file is used.
OpenTracing performs deep copy of arguments passed to the resolver function when args filtering is used (eg. to hide passwords), but this apparently fails there's larger uploaded file in the args.

Potential fix would be default filter that replaces uploaded files with cheap str representation (eg. `<UploadedFile(name="test.jpg", type="image/jpeg", size=44100)>`) before custom filtering logic is ran next.

kuchichan:
Fixes #487 

Additional `TraceableFile` class was add, `__deepcopy__` was overloaded, and right now it returns f-string. Tested for both wsgi and asgi (since multipart form is handled in different way).


ariadne/contrib/tracing/opentracing.py
```diff
@@ -62,6 +66,10 @@ async def resolve(
     def filter_resolver_args(
         self, args: Dict[str, Any], info: GraphQLResolveInfo
     ) -> Dict[str, Any]:
+        for key, val in args.items():
+            if isinstance(val, (cgi.FieldStorage, UploadFile)):
+                args[key] = TraceableFile(val)
```
rafalp, line 71: This should be recursive because GraphQL inputs can be nested.

It also looks like we are modifying `args` in place. We can't do that.

I'm thinking we should implement custom `safe_copy_args` function that would return deep copy of arguments and swap uploaded files with uploaded file's repr. It could default to `str()` for types other than lists/dicts/bool/numeric.

tests/tracing/test_opentracing_sync.py
```diff
@@ -1,5 +1,7 @@
 from unittest.mock import ANY, call
 
+import io
+import cgi
```
rafalp, line 4: Imports should follow PEP-8:

```
import stdlib

import thirdparty

import firstparty
```

rafalp (changes requested):

ariadne/contrib/tracing/opentracing.py
```diff
@@ -65,7 +66,7 @@ def filter_resolver_args(
         if not self._arg_filter:
             return args
```
rafalp, line 67: This value also needs to be made safe. `_arg_filter` is user-provided filter for unsafe args (like passwords), but we also log args when no user filter was provided.

ariadne/contrib/tracing/opentracing.py
```diff
@@ -101,3 +102,33 @@ def opentracing_extension(*, arg_filter: Optional[ArgFilter] = None):
 
 def opentracing_extension_sync(*, arg_filter: Optional[ArgFilter] = None):
     return partial(OpenTracingExtensionSync, arg_filter=arg_filter)
+
+
+def safe_copy_args(args: Dict[str, Any]) -> Dict[str, Any]:
```
rafalp, line 107:
```
def safe_copy_args(value: Any):
    if isinstance(value, dict):
        return {k: safe_copy_args(v) for k, v in value.items()}
    if isinstance(value, list):
        return (safe_copy_args(v) for v in value.items())
    if isinstance(value, (UploadFile, cgi.FieldStorage)):
        return fancy_upload_repr(value)
    return str(value)
```

rafalp (changes requested):

ariadne/contrib/tracing/opentracing.py
kuchichan: Fixed.

ariadne/contrib/tracing/opentracing.py
kuchichan: Fixed.

tests/tracing/test_opentracing_sync.py
```diff
@@ -104,3 +109,67 @@ def test_opentracing_extension_doesnt_break_introspection(schema):
         schema, {"query": introspection_query}, extensions=[OpenTracingExtension]
     )
     assert "errors" not in result
+
+
+def test_arg_filter_resolver_handles_field_storage_with_file(mocker):
```
rafalp, line 114:
```suggestion
def test_resolver_args_filter_handles_uploaded_files_from_wsgi(mocker):
```

tests/tracing/test_opentracing_sync.py
```diff
@@ -104,3 +109,67 @@ def test_opentracing_extension_doesnt_break_introspection(schema):
...
+    info = mocker.Mock()
+
+    copied_kwargs = extension.filter_resolver_args(kwargs, info)
+    assert (
+        f"<class 'cgi.FieldStorage'>"
+        f"(mime_type={field_storage.type}, size={file_size}, filename={field_storage.filename})"
+    ) == copied_kwargs["0"]
+
+
+def test_safe_copy_args():
```
rafalp, line 136:
```suggestion
def test_resolver_args_with_uploaded_files_from_wsgi_are_copied_for_tracing():
```

tests/tracing/test_opentracing.py
```diff
@@ -115,3 +116,22 @@ async def test_opentracing_extension_doesnt_break_introspection(schema):
         schema, {"query": introspection_query}, extensions=[OpenTracingExtension]
     )
     assert "errors" not in result
+
+
+@pytest.mark.asyncio
+async def test_filter_resolver_args_handles_upload_files(mocker):
```
rafalp, line 122:
```suggestion
async def test_resolver_args_filter_handles_uploaded_files_from_asgi(mocker):
```

rafalp (changes requested):

tests/tracing/test_opentracing.py
kuchichan: Done.

tests/tracing/test_opentracing_sync.py
kuchichan: Fixed.

tests/tracing/test_opentracing_sync.py
kuchichan: Fixed.

rafalp (approved):

PR #552, opened 2021-04-23
Defining multiple primary keys in entities 

kuchichan:
Consider this:
``` graphql
type Product @key(fields: "upc") @key(fields: "sku") {
  upc: String!
  sku: String!
  price: String
}
```
Right now (according to tests 🔬  + discussion thread 🧶 ) Ariadne throws an exception: 
`TypeError: The directive 'key' can only be used once at this location.`
It looks like addition of   `repeatable` to the `@key ` directive definition might solve the problem, but further comparison with Apollo-Federation behaviour is needed.


sMorm:
Thanks for getting this up! What is left in to-do in order to get this merged?

sMorm:
hey @kuchichan — any updates on this?

mjfaga:
If it helps, Apollo is building out a test suite to call out federation gaps vs. the spec here: https://github.com/apollographql/apollo-federation-subgraph-compatibility#latest-results

You'll see that multiple `@key` directive support is one of the core pieces of functionality listed.

rafalp:
If this PR fixes the issue, I think it will only need to get out of draft and pass code review.

mjfaga:
@kuchichan can you pull this out of draft so we can iterate forward on it now that we have alignment with the federated spec and pertaining fix implemented in here please?

bekher:
> @kuchichan can you pull this out of draft so we can iterate forward on it now that we have alignment with the federated spec and pertaining fix implemented in here please?

++ It would be great to see this added to Ariadne soon @kuchichan

tests/federation/test_schema.py
```diff
@@ -336,6 +359,73 @@ def user_reference_resolver(_obj, _info, reference):
     assert result.data["_entities"][1]["firstName"] == "James"
 
 
+@pytest.mark.parametrize("primary_key", ["sku", "upc"])
+def test_federated_schema_execute_reference_resolver_with_repeatable_keys(primary_key):
```
rafalp, line 363:
```suggestion
def test_federated_schema_execute_reference_resolver_with_different_keys(primary_key):
```

tests/federation/test_schema.py
```diff
@@ -145,6 +145,29 @@ def test_federated_schema_not_mark_type_with_no_keys():
     assert schema.get_type("_Entity") is None
 
 
+def test_federated_schema_with_repeatable_keys():
```
rafalp, line 148:
```suggestion
def test_federated_schema_with_multiple_keys():
```

PR #565, opened 2021-04-29
[Federation]  Query validation

### LINKED ISSUE #522: Federated schemas should not require at least one query to be implemented
In a Federated environment, the Gateway instantiates the Query type by default. This means that an implementing services should _not_ be required to implement or extend a query. 

# Ideal Scenario
This is an example scenario of what is valid in Node and Java implementations. For example, it should be valid to expose a service that exposes no root queries, but only the federated query fields, like below.

Produced Query type:

**Example**: This is what the schemas would look like for two federated services:
## Product Service
product/schema.gql
```gql
extend type Query {
   products: [Product]
}

type Product {
  id: ID!
  name: String
  reviews: [ProductReview]
}

extend type ProductReview @key(fields: "id") {
  id: ID! @external
}
```
**Output**:

```
products: [Product]
_entities(representations: [_Any]): [_Entity]
_service: _Service
```

## Review Service
review/schema.gql
```gql
# Notice how we don't have to extend the Query type
type ProductReview @key(fields: "id") {
   id: ID!
   comment: String!
}
```

**Output**:
This should be valid.
```
_entities(representations: [_Any]): [_Entity]
_service: _Service
```

# Breaking Scenario
When attempting to implement the `ProductReview` service (see example above) without extending the Query type, Ariadne will fail to [generate a federated schema](https://github.com/mirumee/ariadne/blob/master/ariadne/contrib/federation/schema.py#L57). This is because `make_executable_schema` attempts to generate a federated schema by [extending a Query type](https://github.com/mirumee/ariadne/blob/master/ariadne/contrib/federation/schema.py#L24) with the assumption that a Query type has been defined, which technically it isn't.

kuchichan:
Fixes #522, initally tested with: https://github.com/kuchichan/ariadne-federation-multiple-keys (Query removed from product service)

Todo:
- [x] Write real unit test
- [x] Remove a `repeatable` in key directive (It's a matter of different PR)

### DIFF
--- a/ariadne/contrib/federation/schema.py
+++ b/ariadne/contrib/federation/schema.py
@@ -2,6 +2,7 @@ from typing import Dict, List, Type, Union, cast
 
 from graphql import extend_schema, parse
 from graphql.language import DocumentNode
+from graphql.language.ast import ObjectTypeDefinitionNode
 from graphql.type import (
     GraphQLObjectType,
     GraphQLSchema,
@@ -17,13 +18,13 @@ from .utils import get_entity_types, purge_schema_directives, resolve_entities
 federation_service_type_defs = """
     scalar _Any
 
-    type _Service {
+    type _Service {{
         sdl: String
-    }
+    }}
 
-    extend type Query {
+    {extend_token}type Query {{
         _service: _Service!
-    }
+    }}
 
     directive @external on FIELD_DEFINITION
     directive @requires(fields: String!) on FIELD_DEFINITION
@@ -41,6 +42,17 @@ federation_entity_type_defs = """
 """
 
 
+def has_query_type(type_defs: str) -> bool:
+    ast_document = parse(type_defs)
+    for definition in ast_document.definitions:
+        if (
+            isinstance(definition, ObjectTypeDefinitionNode)
+            and definition.name.value == "Query"
+        ):
+            return True
+    return False
+
+
 def make_federated_schema(
     type_defs: Union[str, List[str]],
     *bindables: Union[SchemaBindable, List[SchemaBindable]],
@@ -52,8 +64,12 @@ def make_federated_schema(
     # Remove custom schema directives (to avoid apollo-gateway crashes).
     # NOTE: This does NOT interfere with ariadne's directives support.
     sdl = purge_schema_directives(type_defs)
+    extend_token = "extend " if has_query_type(sdl) else ""
+    federation_service_type = federation_service_type_defs.format(
+        extend_token=extend_token
+    )
 
-    type_defs = join_type_defs([type_defs, federation_service_type_defs])
+    type_defs = join_type_defs([type_defs, federation_service_type])
     schema = make_executable_schema(
         type_defs,
         *bindables,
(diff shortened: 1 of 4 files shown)


ariadne/contrib/federation/schema.py
```diff
@@ -52,8 +64,12 @@ def make_federated_schema(
     # Remove custom schema directives (to avoid apollo-gateway crashes).
     # NOTE: This does NOT interfere with ariadne's directives support.
     sdl = purge_schema_directives(type_defs)
+    extend_token = "extend " if has_query_type(sdl) else ""
```
rafalp, line 67: I'm thinking if this wouldn't be readable if we inserted `extend type` or `type` into query template.

tests/federation/test_schema.py
```diff
@@ -720,3 +720,28 @@ def test_federated_schema_query_service_ignore_custom_directives():
...
+        name: String
+        price: Int
+        weight: Int
+    }
+    """
+
+    schema = make_federated_schema(type_defs)
+    result = graphql_sync(
+        schema,
+        """query GetServiceDetails {
```
rafalp, line 738:
```suggestion
        """
        query GetServiceDetails {
```

ariadne/contrib/federation/schema.py
kuchichan: Agreed 👌

kuchichan: Done.

tests/federation/test_schema.py
kuchichan: Done.

ariadne/contrib/federation/schema.py
```diff
@@ -17,18 +18,18 @@
...
+    {type_token} Query {{
         _service: _Service!
-    }
+    }}
 
     directive @external on FIELD_DEFINITION
     directive @requires(fields: String!) on FIELD_DEFINITION
     directive @provides(fields: String!) on FIELD_DEFINITION
-    directive @key(fields: String!) on OBJECT | INTERFACE
+    directive @key(fields: String!)  on OBJECT | INTERFACE
```
rafalp, line 32:
```suggestion
    directive @key(fields: String!) on OBJECT | INTERFACE
```

tests/federation/test_schema.py
```diff
@@ -810,3 +810,29 @@ def test_federated_schema_query_service_ignore_custom_directives():
...
+        price: Int
+        weight: Int
+    }
+    """
+
+    schema = make_federated_schema(type_defs)
+    result = graphql_sync(
+        schema,
+        """
+        query GetServiceDetails {
```
rafalp, line 829:
```suggestion
            query GetServiceDetails {
```

rafalp:
@kuchichan LGTM! Could you also add line about this to changelog? Eg.:

"Made `Query` type optional in federated schemas."

salwator (approved):

salwator:
@rafalp maybe we could merge this PR?

rafalp (approved):

rafalp:
@salwator yup, lets merge!

PR #613, opened 2021-07-05
Fix pylint warning

rafalp:
Fixes pylint warning complaining that we are iterating on subscription keys to modify dict's values, eg:

```python
for key in some_dict:
    some_dict[key].do_smth()
```

Where we could do:

```python
for value in some_dict.values():
    value.do_smth()
```


kuchichan (approved):
👌

PR #616, opened 2021-07-05
Regen requirements.txt instead of updating it

rafalp:
Regens `requirements.txt`, effectively updating our deps.


kuchichan (approved):
🏄‍♂️

rafalp:
Starlette 0.15 added quite a few extra deps. I'll let this PR hang for a moment so I can think this through if we want to keep starlette as required dependency.

PR #617, opened 2021-07-05
Added convert_kwargs_to_snake_case_middleware

nilsdet:
This pull request adds a simple middleware, that applies the convert_kwargs_to_snake_case decorator to all resolvers.

### DIFF
--- a/ariadne/middleware.py
+++ b/ariadne/middleware.py
@@ -1,0 +1,17 @@
+from typing import Callable
+
+from graphql import GraphQLResolveInfo
+
+from .utils import convert_kwargs_to_snake_case
+
+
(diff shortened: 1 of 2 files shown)


rafalp:
Hey! Thanks for taking time to open PR with new feature, but we ask that additions to our featureset are put up for discussion first.

GraphQL middleware doesn't feel like right way to go about this problem - they add bunch of extra function calls on top of every resolver call, and those are less than stellar performance wise when result set becomes larger (maybe Py 3.11 will fix this with new call optimizations, but I digress).

Personally I would implement an utility function that takes executable schema returned from `make_executable_schema`, scans it for fields taking arguments, and then if those fields names need conversion, wrap resolver in decorator.

Thank you for your understanding and stepping up to contribute!

PR #640, opened 2021-08-04
Add ability to override `graphql` arguments from ASGI and WSGI apps

danields761:
Add ability to override `graphql` arguments from ASGI and WSGI apps
and remove hard-coded `execution_context_class` from `ariadne.graphql.graphql`
and `ariadne.graphql.graphql_sync` functions.

The motivation for this change is to give a user ability to customize
GraphQL behaviour in various ways. As an example user can override error
handling mechanism completely via custom `execution_context_class` or
pass custom field resolver etc.

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -89,6 +89,7 @@ class GraphQL:
         extensions: Optional[Extensions] = None,
         middleware: Optional[Middlewares] = None,
         keepalive: float = None,
+        **graphql_kwargs: Any,
     ):
         self.context_value = context_value
         self.root_value = root_value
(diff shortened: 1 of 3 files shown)


ariadne/asgi.py
```diff
@@ -89,6 +89,7 @@ def __init__(
         extensions: Optional[Extensions] = None,
         middleware: Optional[Middlewares] = None,
         keepalive: float = None,
+        **graphql_kwargs: Any,
```
rafalp, line 92: This will make any typos in options be eaten up and passed all the way down to `graphql()` call. We should be explicit in what options we take.

rafalp (changes requested):

rafalp:
Closing due to inactivity. If somebody wants to contribute this, please add `graphql(...)` options as explicit kwargs on `GraphQL(...)`. Thanks!

PR #650, opened 2021-08-20
RFC: Configurable JSON response class

duckontheweb:
## Context

We are working on a FastAPI app that uses Ariadne's ASGI server for its GraphQL endpoints. We are using `ORJSONResponse` throughout that app and would like to be able to use the same response class for our GraphQL requests. Performance is definitely a big motivator for this, but we also want to ensure that response formats are consistent throughout the API, and using the same JSON serializer throughout is a good step in that direction.

## Changes

This PR adds a `json_response_class` parameter to `ariadne.asgi.GraphQL.__init__` that may be any class inheriting from `starlette.responses.JSONResponse` defaults to that class. This value of this argument is then used as the response class in `GraphQL.graphql_http_server`.

I took a look at the websocket handlers in that class and did not see any place where this change should apply, but I missed something I'm certainly happy to correct it. If this changes ends up being accepted, I'm also happy to put in an equivalent PR on `ariadne-website` to update the docs.

### DIFF
--- a/ariadne/asgi.py
+++ b/ariadne/asgi.py
@@ -90,7 +90,6 @@ class GraphQL:
         extensions: Optional[Extensions] = None,
         middleware: Optional[Middlewares] = None,
         keepalive: float = None,
-        json_response_class: Type[JSONResponse] = JSONResponse,
     ):
         self.context_value = context_value
         self.root_value = root_value
@@ -105,7 +104,6 @@ class GraphQL:
         self.middleware = middleware
         self.keepalive = keepalive
         self.schema = schema
-        self.json_response_class = json_response_class
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send):
         if scope["type"] == "http":
@@ -170,6 +168,15 @@ class GraphQL:
     ) -> Response:
         return HTMLResponse(PLAYGROUND_HTML)
 
+    def build_json_response(
+        self,
+        content: Dict[str, Any],
+        status_code: int
+    ) -> Type[JSONResponse]:
+        """Override this method to use a different response class or to otherwise change the way the JSON HTTP response
+        is constructed."""
+        return JSONResponse(content, status_code=status_code)
+
     async def graphql_http_server(self, request: Request) -> Response:
         try:
             data = await self.extract_data_from_request(request)
@@ -194,7 +201,7 @@ class GraphQL:
             middleware=middleware,
         )
         status_code = 200 if success else 400
-        return self.json_response_class(response, status_code=status_code)
+        return self.build_json_response(response, status_code=status_code)
 
     async def extract_data_from_request(self, request: Request):
         content_type = request.headers.get("Content-Type", "")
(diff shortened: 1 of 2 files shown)


duckontheweb:
Just for some additional context, we are currently patching this by sub-classing the `GraphQL` class and overwriting the `graphql_http_server` method like so:

```python
from ariadne.exceptions import HttpError
from ariadne.graphql import graphql
from fastapi.responses import ORJSONResponse
from starlette.requests import Request
from starlette.responses import PlainTextResponse, Response


class OrjsonGraphQL(GraphQL):
    async def graphql_http_server(self, request: Request) -> Response:
        try:
            data = await self.extract_data_from_request(request)
        except HttpError as error:
            return PlainTextResponse(error.message or error.status, status_code=400)

        context_value = await self.get_context_for_request(request)
        extensions = await self.get_extensions_for_request(request, context_value)
        middleware = await self.get_middleware_for_request(request, context_value)

        success, response = await graphql(
            self.schema,
            data,
            context_value=context_value,
            root_value=self.root_value,
            validation_rules=self.validation_rules,
            debug=self.debug,
            introspection=self.introspection,
            logger=self.logger,
            error_formatter=self.error_formatter,
            extensions=extensions,
            middleware=middleware,
        )
        status_code = 200 if success else 400
        return ORJSONResponse(response, status_code=status_code)
```

rafalp:
To be honest I would've preferred for `GraphQL`  to be updated so there was `return build_json_response(request, success, response)` method on `GraphQL` for devs to override through inheritance instead.

Seeing how many people are already using Ariadne together with FastAPI I wouldn't even mind if default implementation for `ASGI` attempted to import `ORJSONResponse` and only fallbacked to Starlette's JSON response.

ariadne/asgi.py
```diff
@@ -167,6 +168,15 @@ async def render_playground(  # pylint: disable=unused-argument
     ) -> Response:
         return HTMLResponse(PLAYGROUND_HTML)
 
+    def build_json_response(
```
rafalp, line 171: Please move this under `extract_data_from_multipart_request` and make it take `request: Request` as first argument.

ariadne/asgi.py
```diff
@@ -167,6 +168,15 @@ async def render_playground(  # pylint: disable=unused-argument
...
+        self,
+        content: Dict[str, Any],
+        status_code: int
+    ) -> Type[JSONResponse]:
```
rafalp, line 175:
```suggestion
    ) -> Type[Response]:
```

Not that correct but makes possible inheritance chain more forgiving.

ariadne/asgi.py
```diff
@@ -191,7 +201,7 @@ async def graphql_http_server(self, request: Request) -> Response:
             middleware=middleware,
         )
         status_code = 200 if success else 400
-        return JSONResponse(response, status_code=status_code)
+        return self.build_json_response(response, status_code=status_code)
```
rafalp, line 204:
```suggestion
        return self.build_json_response(request, response, status_code)
```

rafalp (changes requested):

rafalp:
We've made this easier to customize via inheritance in #768. Thanks for pointing out need for this!

PR #657, opened 2021-08-27
Benchmark

kubaklimas:
Creating json files with data for benchmark example app. 
Measure time of post to localhost with simple and complex query. 
To start comparing execution times all you need to do start tests in folder benchmarks.
It will show exectution times(min/max/avg) 
To refresh json files you need to run script ` json_data.py`


benchmarks/conftest.py
```diff
@@ -1,9 +1,24 @@
+from attr import dataclass
 import json
 import pytest
 
 from ariadne import QueryType, gql, make_executable_schema
 
 
+@dataclass
+class Person:
+    name = "Name"
```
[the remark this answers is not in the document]
kubaklimas: Done, now data for dataclasses are loaded from Json files

benchmarks/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+class Person:
+    name: str
+    age: int
+
+
+@dataclass
+class User:
+    id: int
+    name: str
+    group: dict
```
rafalp, line 18: Group should be other (nested) Dataclass

benchmarks/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+    avatar: list
```
rafalp, line 19: This should be `List[Avatar]` where `Avatar` is dataclass.

benchmarks/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+@pytest.fixture
+def complex_data_from_json():
+    with open("benchmarks/complex.json", "r") as file:
+        return json.load(file)
+
+
+@pytest.fixture
+def complex_query():
```
rafalp, line 111: This can be a string in tests file.

benchmarks/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+    person = simple_data[0]
+    return {"name": person["name"], "age": person["age"]}
+
+
+def user_data_dictionary():
+    user = complex_data[0]
+    return {"name": user["name"], "group": user["group"], "avatar": user["avatar"]}
+
+
+query = QueryType()
```
rafalp, line 39: Building on above ideas (two sets of data, one with dicts and other with objects) we could drop need for custom resolvers and executable schema fixtures.

benchmarks/json_data.py
```diff
@@ -0,0 +1,43 @@
+import json
```
rafalp, line 1: Rename this Python file to `generate_data.py`.

benchmarks/json_data.py
```diff
@@ -0,0 +1,43 @@
...
+import random
+import string
+
```
rafalp, line 4: Unnecessary blank line.

benchmarks/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+    name: str
+    group: dict
+    avatar: list
+
+
+with open("benchmarks/simple.json") as f:
+    simple_data = json.load(f)
+
+with open("benchmarks/complex.json") as f:
+    complex_data = json.load(f)
```
rafalp, line 26: I'm wondering if it wouldn't be simpler to generate single `data.json` file that contains data for entire schema and then create two sets of data from it:

- `raw_data` which contains dicts
- `hydrated_data` which contains objects

benchmarks/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+            % name
+        )
+        return query
+
+    return name_of_query
+
+
+@pytest.fixture
+def simple_query():
```
rafalp, line 137: Same as above

benchmarks/schema.gql
```diff
@@ -0,0 +1,42 @@
...
+        people: [Person!]!
+        person_dataclass: [Person!]!
+        person: [Person!]!
+        users_dataclass: [User!]
+        users: [User!]
+        user_dataclass: [User!]
+        user: [User!]
+    }
+
+    type Person {
```
rafalp, line 12: Let's drop this type and query for  scalar `User` fields in "simple" test cases.

benchmarks/test_complex.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_benchmark_complex_query_post_return_list_of_500_elements(
+    benchmark, complex_query, complex_data_from_json, schema
+):
+    app = GraphQL(schema, root_value=complex_data_from_json)
+    client = TestClient(app)
+
+    def wrapper():
+        request = client.post("/", json={"query": complex_query("users")})
+        return request
```
rafalp, line 14: Let's do `return client...`

benchmarks/test_complex.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_benchmark_complex_query_post_return_one_element(
+    benchmark, complex_query, schema
+):
+    app = GraphQL(schema)
+    client = TestClient(app)
+
+    def wrapper():
+        request = client.post("/", json={"query": complex_query("user")})
+        return request
```
rafalp, line 29: Same here.

benchmarks/test_complex.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_benchmark_complex_query_post_return_list_500_dataclass(
+    benchmark, complex_query, schema
+):
+    app = GraphQL(schema)
+    client = TestClient(app)
+
+    def wrapper():
+        request = client.post("/", json={"query": complex_query("users_dataclass")})
+        return request
```
rafalp, line 44: Ditto

benchmarks/test_complex.py
```diff
@@ -0,0 +1,63 @@
...
+    result = benchmark(wrapper)
+
+    assert result.status_code == 200
+
+
+def test_benchmark_complex_query_post_return_dataclass(
+    benchmark, complex_query, schema
+):
+    app = GraphQL(schema)
+    client = TestClient(app)
```
rafalp, line 55: Ditto

benchmarks/schema.gql
```diff
@@ -0,0 +1,42 @@
+type Query {
+        people_dataclass: [Person!]!
```
rafalp, line 2: Don't use spearate `_dataclass` fields. We can use `root_value` to pass either raw or hydrated data to `GraphQL()` app for default resolvers to use.

rafalp (changes requested):

benchmarks/conftest.py
kubaklimas: Now group and avatar are other Dataclasses

benchmarks/conftest.py
kubaklimas: Same as above

benchmarks/conftest.py
kubaklimas: Done, now every test have own query inside of it

benchmarks/conftest.py
kubaklimas: We are using only default resolvers now

benchmarks/json_data.py
kubaklimas: Done

benchmarks/json_data.py
kubaklimas: Removed

benchmarks/conftest.py
kubaklimas: Now we have only one file data.json

benchmarks/conftest.py
kubaklimas: Done

benchmarks/schema.gql
kubaklimas: Removed, now we only have User

benchmarks/test_complex.py
kubaklimas: Done

benchmarks/test_complex.py
kubaklimas: done

benchmarks/test_complex.py
kubaklimas: done

benchmarks/schema.gql
kubaklimas: As above, we have only one query now and we pass data by root_value

benchmarks/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+    for user in users_data():
+        return [user]
+
+
+query = QueryType()
+
+
+@pytest.fixture
+def type_defs():
+    with open("benchmarks/schema.gql", "r") as file:
```
rafalp, line 64: Ariadne has build-in function for this: [`load_schema_from_path`](https://ariadnegraphql.org/docs/api-reference#load_schema_from_path)

benchmarks/schema.gql
```diff
@@ -0,0 +1,29 @@
+type Query {
+        users:[User!]
```
rafalp, line 2: Please format this file with some GraphQL formatter.

benchmarks/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+    avatar: List[Avatar]
+
+
+with open("benchmarks/data.json") as f:
+    data = json.load(f)
+
+
+def users_data():
+    for row in data["users"]:
+        yield User(**row)
```
rafalp, line 35: Remember that `dataclass` creation is not recursive. You need to manually hydrate Group and Avatars.

rafalp (changes requested):

benchmarks/conftest.py
kubaklimas: Replaced with load_schema_from_path function

benchmarks/schema.gql
kubaklimas: Done

benchmarks/conftest.py
kubaklimas: Done, used "get" instead of "pop". Pop worked only for first test ( no matter which were situated in file as first)

benchmarks/conftest.py
```diff
@@ -0,0 +1,72 @@
+from attr import dataclass
```
rafalp, line 1: Move this line before `from typing import List`

rafalp:
This is looking quite good now. I would like to run this locally to see how it works before making final call but I'll need to find time for that, sorry. :|

rafalp (approved):

PR #661, opened 2021-09-01
Fix "isinstance" across default resolvers

### LINKED ISSUE #660: snake_case_fallback_resolvers not calling obj.get(attr_name)
**Ariadne version:** 0.13.0
**Python version:** 3.8.11

Hello. I am using the [databases](https://www.encode.io/databases/) package with an [asyncpg](https://magicstack.github.io/asyncpg/current/) backend to interact with a PostgreSQL database. The objects returned from my queries are of the type `databases.backends.postgres.Record`.  The desired attributes can only can accessed via the get method. However, when I use `snake_case_fallback_resolvers`, Ariadne has trouble resolving the requested fields and I receive the following error: `Cannot return null for non-nullable field`

If I instead use the regular `fallback_resolvers` (adjusting my schema's naming conventions), Ariadne is able to resolve the requested fields.

Is this a bug or am I doing something wrong? Thank you for your time.

kuchichan:
Fixes #660

### DIFF
--- a/tests/test_kwargs_camel_case_conversion.py
+++ b/tests/test_kwargs_camel_case_conversion.py
@@ -5,14 +5,14 @@ from ariadne import convert_kwargs_to_snake_case
 
 def test_decorator_converts_kwargs_to_camel_case():
     @convert_kwargs_to_snake_case
-    def my_func(*_, **kwargs):
+    def wrapped_func(*_, **kwargs):
         assert kwargs == {
             "first_parameter": True,
             "second_parameter": "value",
             "nested_parameter": {"first_sub_entry": 1, "second_sub_entry": 2},
         }
 
-    my_func(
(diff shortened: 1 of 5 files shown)


tests/test_kwargs_camel_case_conversion.py
```diff
@@ -51,6 +67,22 @@ def my_func(*_, **kwargs):
     )
 
 
+def test_decorator_converts_objects_in_lists_to_camel_case_for_mapping(fake_mapping):
```
rafalp, line 70:
```suggestion
def test_decorator_converts_mappings_in_lists_to_camel_case(fake_mapping):
```

kuchichan: Done 👍

rafalp (approved):

PR #711, opened 2021-10-27
Handle possible None enum value

kissgyorgy:
`default_value` can be None in `InputValueDefinitionNode` (and any other subclass
of `DefinitionNode.default_value` property), so it should be handled at all times.

Refactored the `enum_values_in_input_type` and `enum_values_in_field_args` to use
the same helper function, so they can consistently handle the same edge case.

### DIFF
--- a/ariadne/enums.py
+++ b/ariadne/enums.py
@@ -159,16 +159,7 @@ def enum_values_in_input_type(
     type_: GraphQLInputObjectType,
     field_name,
 ) -> Generator[InputFieldWithKeys, None, None]:
-    for input_name, field in type_.fields.items():
-        resolved_type = resolve_null_type(field.type)
-        if isinstance(resolved_type, GraphQLEnumType):
-            yield field_name, input_name, field, None
-
-        if isinstance(resolved_type, GraphQLInputObjectType):
-            if field.ast_node is not None:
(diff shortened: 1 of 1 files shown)


rafalp:
Hi! Can you add tests demonstrating the issue fixed here?

ariadne/enums.py
```diff
@@ -154,23 +154,27 @@ def enum_values_in_object_type(
         yield from enum_values_in_field_args(field_name, field)
 
 
-@enum_values_in_types.register(GraphQLInputObjectType)
-def enum_values_in_input_type(
-    type_: GraphQLInputObjectType,
-    field_name,
-) -> Generator[InputFieldWithKeys, None, None]:
-    for input_name, field in type_.fields.items():
+def _get_field_with_keys(field_name, fields):
```
rafalp, line 157: We are ordering logic in our code from top to bottom, so this should be declared below `enum_values_in_field_args`.

rafalp (changes requested):

kissgyorgy:
> Hi! Can you add tests demonstrating the issue fixed here?

Yes, sorry. I will add a minimal reproduction about the issue.

rafalp:
I'll tangentially schedule this on next release. Sounds like helpful thing to fix.

rafalp (approved):

PR #765, opened 2022-01-12
allow tuples along with lists for isinstance()

dand-oss:
Tuples are treated like dicts, because isinstance() use looks only for lists

### DIFF
--- a/ariadne/contrib/federation/schema.py
+++ b/ariadne/contrib/federation/schema.py
@@ -58,7 +58,7 @@ def make_federated_schema(
     *bindables: Union[SchemaBindable, List[SchemaBindable]],
     directives: Dict[str, Type[SchemaDirectiveVisitor]] = None,
 ) -> GraphQLSchema:
-    if isinstance(type_defs, list):
+    if isinstance(type_defs, (list, tuple)):
         type_defs = join_type_defs(type_defs)
 
     # Remove custom schema directives (to avoid apollo-gateway crashes).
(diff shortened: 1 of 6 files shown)


rafalp:
Hey, thanks for PR. I see it changes quite a few places. Can you say if you've verified that every of those type checks can actually receive tuple?

dand-oss:
thanks for the feedback - GOOD maintenance quetions.

Possible answers

1.  If you have unit tests for this areas, I can add some test which prove tuple.
2.  I can just run the system a while and report back - and visually inspect use of list vs tuple in code areas.
3. I can drop tuple fixes and limit fix to the single callback case that blows up on our application because mypy suggested using tuples instead of lists for certain cases.
4. We could pick a protocol for RTTI type branching instead of enumerating - currently is "if not a list MUST be a dict"

Please advise?

rafalp:
@dand-oss Hey, sorry for late response.

My problem with this PR is that it looks like somebody did find and replace `isinstance(*, list)` to `isinstance(*, (list, tuple))` without checking if either still supporting `list` or supporting both makes sense.

rafalp:
I understand now that in context of supporting both old and new GraphQL core this PR made sense. 😅

TBH I'm in favor of single PR that just updates graphql-core version to 3.2 and drops support for 3.1x I think we will do it in 0.15.x release for which I have one extra revolution planned.

rafalp:
Supersed by #796

PR #783, opened 2022-01-25
Allow make_schema to take directives again. Fixes #780

### LINKED ISSUE #780: graphql-core v3.2 breaks directives
Since [this commit](https://github.com/graphql-python/graphql-core/commit/b9423b74ca9d22d009a1a4633b38b2bcd06c604b) in graphql-core replaced all lists with tuples, `schema.directives` is a tuple now.

This causes an error in [ariadne.schema_visitor.each](https://github.com/mirumee/ariadne/blob/3921fb1d7cd156eb7b5dc39e9a0695508f187cdd/ariadne/schema_visitor.py#L58).

It seems like a possible solution is to replace `isinstance(list_or_dict, list)` with `isinstance(list_or_dict, Sequence)`.

joidegn:
Fix #780 using @sda97ghb's idea while we are waiting for https://github.com/mirumee/ariadne/pull/765 to be tested.

I tested this in our setup and it seems to work.

### DIFF
--- a/ariadne/schema_visitor.py
+++ b/ariadne/schema_visitor.py
@@ -11,6 +11,7 @@ from typing import (
     Union,
     cast,
 )
+from collections.abc import Sequence
 
 from graphql import is_named_type, value_from_ast_untyped
 from graphql.execution.values import get_argument_values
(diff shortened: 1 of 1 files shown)


rafalp:
Supersed by #796

PR #824, opened 2022-03-22
Use `execute_sync` instead of `execute` in `graphql_sync`

### LINKED ISSUE #823: Use `execute_sync` in `graphql_sync`
`graphql_sync` should use `execute_sync` instead of `execute`.

I run my code in a synchronous context, but a resolver was identified as asynchronous due to the fact that GraphQL core does a simple `hasattr(value, "__await__")` check instead of using `inspect.isawaitable`. This caused a false positive.

However, when using `graphql_sync` I expect that this check does not even happen. Consider using [`execute_sync`](https://github.com/graphql-python/graphql-core/blob/27768d3faac5bb8ebb2c6a151a7192b7fa498f2f/src/graphql/execution/execute.py#L1061) instead.

louwers:
Fixes #823

### DIFF
--- a/ariadne/graphql.py
+++ b/ariadne/graphql.py
@@ -20,6 +20,7 @@ from graphql import (
     GraphQLSchema,
     TypeInfo,
     execute,
+    execute_sync,
     parse,
     subscribe as _subscribe,
 )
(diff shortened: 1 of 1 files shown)


tmc:
@rafalp PTAL.

rafalp:
This is on the list to check for 0.15 release.

rafalp (approved):

rafalp:
For some reason GitHub decided it won't be running PR checks for other people, but I've checked this branch locally and tests are passing.

rafalp:
Thank you!

PR #828, opened 2022-03-31
Support isinstance() checks in the base types

vmarkovtsev:
To avoid the error on e.g. `isinstance(QueryType(), SchemaBindable)`

```
TypeError: Instance and class checks can only be used with @runtime_checkable protocols
```


rafalp (changes requested):

rafalp:
@vmarkovtsev we also need to to move extension classes to protocol.

vmarkovtsev:
@rafalp why do we need to move classes in the same PR? Or I don't get what you request...

rafalp:
@vmarkovtsev can you rebase against master so tests pipeline is run? You will see an error then that `@runtime_checkable` can only be used on classes that extend `Protocol` which is not the case for `Extension`s.

vmarkovtsev:
I've rebased, however, there are two things:
1. `Extension` does inherit from `Protocol`
image
2. CI does not trigger for first-time contributors like me and has to be approved manually, I think.

rafalp:
Okay, so GH is not letting me run CI for this PR because for some reason there's no "Approve and Run" option 🤦‍♂️

rafalp:
@vmarkovtsev and it failed: https://github.com/mirumee/ariadne/runs/5989948597?check_suite_focus=true

vmarkovtsev:
LOL. I'll try to PR a fix ASAP!

rafalp:
@vmarkovtsev I've removed decorator from one extension class for now and I am researching why GitHub is not running actions for forks. Looks like PR authors need to enable actions on their fork first?

rafalp:
I'll release current master as 0.15.0dev5 so you can ride the edge if you want to.

vmarkovtsev:
So
```python
@runtime_checkable
class Extension(Protocol)
```
works but
```python
@runtime_checkable
class ExtensionSync(Extension):
```
doesn't :thinking: Maybe it's not needed because the parent class is already `runtime_checkable` and the reported error is misleading :thinking:

vmarkovtsev:
Thanks @rafalp!

Issue #775, opened 2022-01-19
Federation Support - Description of custom directives breaks make_federated_schema


j-riebe:
I stumpled into a problem with the federation support while using custom directives.
It appears that `make_federated_schema` doesn't support custom directives **if they have a description**.

The **culprit** is following function:
https://github.com/mirumee/ariadne/blob/ceee00cdd332b0ef8165480183516cf201f2e0bc/ariadne/contrib/federation/utils.py#L54-L61

It purges the directives from the schema but keeps the description in place.
This may lead to a violation of the GraphQL Syntax depending on which type definition come next.

The **solution** would be to extend the regex used to capture the directive definition so that it includes block-strings and single-line-strings.

https://github.com/mirumee/ariadne/blob/ceee00cdd332b0ef8165480183516cf201f2e0bc/ariadne/contrib/federation/utils.py#L17-L32

j-riebe:
Thx @rafalp

vikramsubramanian:
It feels like this Regex change would make parsing comments take forever (exponential?)

j-riebe:
What do you mean by _would_ and _it feels_?

Are you experiencing "forever"? 😄

vikramsubramanian:
Hey, sorry for the vagueness. 
Every `"""..."""` comment block seems to exponentially grow the time it takes to parse a Graphql block. Simply removing all my comments compiled my code instantly whereas with ~10 comments, the script took >30 mins and ~30s with 1 comment block

j-riebe:
I'd suggest to open a new issue, that targets the performance issue (as this issue is closed).

Some closing thoughts:
Keep in mind that the original problem is rooted in the use of `purge_schema_directives` which is reused in `make_federated_schema`
https://github.com/mirumee/ariadne/blob/f37df147228f87a356690d2bb9b99a5c1cfa7c63/ariadne/contrib/federation/utils.py#L58-L65
https://github.com/mirumee/ariadne/blob/f37df147228f87a356690d2bb9b99a5c1cfa7c63/ariadne/contrib/federation/schema.py#L64-L68

Maybe the true solution lies in removing `purge_schema_directives` entirely, as its only usage comes with a hacky comment:
> (to avoid apollo-gateway crashes)

If one finds another way `to avoid apollo-gateway crashes`,  we don't need that regex.

That being said, #736 also mentions this specific function. Maybe you can find some help there.

PR #847, opened 2022-04-19
Directive definition regex pref

Mainian:
#775 Introduced a performance issue: https://github.com/mirumee/ariadne/issues/775#issuecomment-1102335619

The solution is to add start and end anchors to the comment regexes.


rafalp (approved):

rafalp:
Thank you!

rafalp:
Gosh, I've missed CI not running by GH again. I'm at lost of words on their changes for this :/

PR #849, opened 2022-04-20
Federated graph regex performance. 

Mainian:
The main performance problem comes from `(?:\"{3}(?:[^\"]|[\s])*?\"{3})`.

The lazy '*?' drastically kills the performance of any federated graph that uses block comments, even if that graph is valid and has no custom directives.

Additionally, the old solution creates a run away regex with something like this: (4 quotes)
```
        """"
        Any Description
        """
        directive @custom on FIELD
        
        type Query {
            rootField: String @custom
        }
```

New solution will just ignore this comment.
```
        """"
        Any Description
        """
        
        type Query {
            rootField: String @custom
        }
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,9 +4,7 @@ on:
   push:
     branches:
       - master
-  pull_request_target:
-    branches:
-      - master
+  pull_request:
 
 jobs:
   build:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - master
   pull_request_target:
+    branches:
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


Mainian:
@rafalp 
`pull_request_target` runs against the workflow and code from the 'base' pull request instead of from the merge commit.

rafalp (approved):

rafalp:
I see, but it was also supposed to run checks on PR's made by first time contributors? 🤔 

I've used `pull_request` previously and it didn't run checks at all. 🤨

Either way this is looking good. Thanks!

Mainian:
@rafalp 
15.0 is basically unusable if you have any multi-line comments in your `types_defs`. 

Does if make sense to push this as a bug fix?

rafalp:
I guess we can release this as 0.15.1 tomorrow. :)

rafalp:
I've had some delay with this but I'll release this today.

rafalp:
@Mainian released!

Issue #837, opened 2022-04-11
Implement graphql-ws protocol support


rafalp:
Currently Ariadne implements `subscription-transport-ws` protocol which Apollo has deprecated and replaced with [graphql-ws](https://github.com/enisdenjo/graphql-ws).

We should either update `ariadne.asgi.GraphQL` to graphql-ws or refactor it to make WebSocket support handled via strategy pattern, eg:

```python
app = GraphQL(schema, websocket=GraphQLWS(options))
```

eugeneyalansky:
There is a need for me to implement. @rafalp Would you like to share some thoughts so we can discuss my contribution?

rafalp:
Hey @eugeneyalansky. I am open to making Ariadne support websocket protocols through strategy pattern as described in first post. So you could implement it yourself and then pass it to Ariadne.

eugeneyalansky:
I'm thinking of subprotocols strategies. 
I'd like also to cover queries and mutations based on WS transport, I've checked the source code on related PRs. 

The protocol specifies how to handle queries and mutations, but on the other hand, distinguishing it from subscriptions without parsing the query into a document seems challenging.

rafalp:
I think we can assume here that if connection was made over the websocket, its going to be subcription. Ariadne's `subscribe` method will do the checking if thats indeed a fact for you.

pigletto:
@rafalp @eugeneyalansky In fact I've already started working on adding graphql-transport-ws protocol (this is the correct protocol name, while project is graphql-ws) to Ariadne. See my fork: https://github.com/pigletto/ariadne/tree/graphql-transport-ws-protocol

Some details about this implementation: The protocol detection is done automatically. Running queries and mutations is also done (for both protocols). I consider this implementation almost done - currently I'm writing tests and doing some minor fixes. Some ideas are borrowed from Strawberry project

eugeneyalansky:
@pigletto What a save, checking out your changes

It's kind of confusing namespacing 😑

graphql-ws = graphql_transport_ws
subscriptions-transport-ws = graphql-ws

rafalp:
@pigletto this is amazing work!

My only suggestion is to make protocols provided to `asgi.GraphQL` through construction options so they are easy to customize by devs having non-standard use cases. Making API easy for customization if you need something Ariadne doesn't provide is one of design paradigms we are following here.

Also would you be ok with me assigning this issue to you?

pigletto:
[...]

PR #855, opened 2022-04-25
Handle graphql-transport-ws protocol (#837)

pigletto:


ariadne/asgi/graphql.py
```diff
@@ -0,0 +1,129 @@
...
+                GraphQLWS.PROTOCOL: GraphQLWS,
+            }
+        else:
+            self.subscription_handlers = {
+                handler.PROTOCOL: handler for handler in subscription_handlers
+            }
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        if scope["type"] == "http":
+            await GraphQLHTTP(
```
rafalp, line 77: Please factor this branch into separate method.

We need to be able to provide custom `GraphQLHTTP` implementation through option.

ariadne/asgi/graphql.py
```diff
@@ -0,0 +1,129 @@
...
+                validation_rules=self.validation_rules,
+                debug=self.debug,
+                introspection=self.introspection,
+                logger=self.logger,
+                error_formatter=self.error_formatter,
+                extensions=self.extensions,
+                middleware=self.middleware,
+            ).handle_http(scope=scope, receive=receive, send=send)
+        elif scope["type"] == "websocket":
+            ws = WebSocket(scope=scope, receive=receive, send=send)
```
rafalp, line 90: This also also should be separate method.

ariadne/asgi/handlers/__init__.py
```diff
@@ -0,0 +1,13 @@
+from .graphql_transport_ws import GraphQLTransportWS
+from .graphql_ws import GraphQLWS
```
rafalp, line 2: I don't think we need to prefix all modules with `graphql`.

Handlers should have `Handler` suffix, eg. `GraphQLHTTPHandler`.

rafalp (changes requested):
Please reorder and format imports and `__all__` lists alphabetically following the pep404 convention of splitting stdlib/3rdparty/1stparty imports with blank lines.

ariadne/asgi/handlers/__init__.py
pigletto: Module names are meant to reflect the protocol names, as the protocols are: "graphql-transport-ws" and "graphql-ws". Does it work for you?

ariadne/asgi/graphql.py
pigletto: The way it is done now makes the code backwards compatible but less elastic indeed. I can refactor this to something like below:

`http_handler = GraphQLHttpHandler(schema, error_formatter=error_formatter)`

`subscription_handler1 = GraphQLWSHandler(schema, error_formatter=error_formatter, http_handler=http_handler)`
`subscription_handler2 = GraphQLTransportWSHandler(schema, error_formatter=error_formatter, http_handler=http_handler)`

`app1 = GraphQL(http_handler=http_handler, subscription_handler=subscription_handler1)`
`app2 = GraphQL(http_handler=http_handler, subscription_handler=subscription_handler2)`

The code is much simpler but it is a breaking change. What do you think?

Extra note: passing **http_handler** to, eg. **GraphQLWSHandler** as a parameter will make it possible to handle queries/mutations over websocket.

pigletto: I've refactored the code according to the proposal above. I've not added the default HTTP handler though. This would require passing all the parameters needed by the GraphQLHttpHandler to the GraphQL constructor.

With explicit handlers:
  
```
http_handler = GraphQLHttpHandler(schema, error_formatter=error_formatter)
app1 = GraphQL(http_handler=http_handler, subscription_handler=subscription_handler1)
```

with default HTTP handler created by GraphQL class (it is not implemented this way) this would be:

`app1 = GraphQL(schema=schema, error_formatter=error_formatter, subscription_handler=subscription_handler1)`

We would possibly need something like the below to implement the above:

```
def __init__(self, *args, **kwargs):
     self.http_handler = kwargs['http_handler'] if 'http_handler' in kwargs else GraphQLHttpHandler(*args, **kwargs)
     self.subscription_handler = kwargs['subscription_handler'] if 'subscription_handler' in kwargs else GraphQLWsHandler(*args, **kwargs)
```

It seems to me that a more explicit way of passing handlers and parameters is better so I've not implemented the above.

rafalp:
This is looking very good *but* I am contemplating two things:

1. Renaming `subscription_handler` to `websocket_handler` because that's what it really is in logic.
2. Keeping options on `asgi.GraphQL` and passing them to handlers in `GraphQL.__init__` via method like `configure()`
3. Setting default HTTP handler if none was provided.

Something like this:

```python
class GraphQL:
    def __init__(self, schema, http_handler, debug):
        if http_handler:
            self.http_handler = http_handler
        else:
            self.http_handler = GraphQLHttpHandler()
            
        self.http_handler.configure(schema=schema, debug=debug)
```

I am just not a big fan of having to specify config twice when my app uses both HTTP and Websockets for GraphQL.

pigletto:
@rafalp I'm a bit lost regarding the configuration specification. 

Currently we have two handlers (one for HTTP and one for websockets) to be configured for GraphQL class. Unfortunately, each implementation of such handlers can have a different set of configuration options. HTTP handler has some parameters that are not used by Websocket handlers but Websocket handlers also have different parameters (graphql-ws handler needs **keepalive** but graphql-transport-ws has **connection_init_wait_timeout**. I can imagine that custom handler implementations can expect other parameters). 
Because of that, we would need to have something I've proposed in my previous comment as explicitly listing all possible __init__ parameters is not possible:

```
def __init__(self, *args, **kwargs):
     self.http_handler = kwargs['http_handler'] if 'http_handler' in kwargs else GraphQLHttpHandler(*args, **kwargs)
     self.subscription_handler = kwargs['subscription_handler'] if 'subscription_handler' in kwargs else GraphQLWsHandler(*args, **kwargs)
```

rafalp:
My idea is that handlers will take their configuration options (eg. `enable_introspection=True: bool` on HTTP handler) but basic settings should be given once to `GraphQL` class and then propagated for you automagically.

pigletto:
@rafalp ok, got the idea now. I've updated the code so that it uses configure method now. I've left the default websocket_handler to be GraphQLWS as this is more backwards compatible but we might change this and require users to define it explicitly.

rafalp:
Looking good, but I don't know why you've changed imports to absolute while we use relative :D

pigletto:
I think I've found the absolute imports in the tests and that's why I've changed this... My bad.

rafalp:
Yup, tests are using absolute imports cos they are in separate package, but package itself uses rellatives.

rafalp:
Just an update that I haven't forgotten about this PR, but I am currently occupied with other tasks at Mirumee that are keeping me away from trying out this PR myself and making final call on merging. But I still like it and want to make it focus on next feature release.

kevinvalk:
Thanks @rafalp and @pigletto! I am actually facing this as we speak, our frontend had some troubles with the unmaintained subscriptions-transport-ws (project) library. Hence would be awesome if Ariadne supports graphql-ws (project). That will make  fixing the frontend a lot easier in our case. Hope you can merge and release soon (any ETA would be much appreciated but I understand if you cannot do so)!

rafalp:
Hey @kevinvalk, I understand your predicament but I can't provide an ETA now and can tell its going to be at least few weeks before I will be able to. :/

rafalp (approved):

rafalp:
The time is now! Thanks for contributing! I'll try to write documentation for this and get this released once the docs are done.

rafalp:
This is now available in `0.16.0b1` beta release.

ariadne/asgi/handlers/graphql_ws.py
```diff
@@ -0,0 +1,293 @@
...
+                try:
+                    result = self.on_operation(websocket, operations[operation_id])
+                    if result and isawaitable(result):
+                        await result
+                except Exception as error:
+                    if not isinstance(error, GraphQLError):
+                        error = GraphQLError(str(error), original_error=error)
+                    log_error(error, self.logger)
+
+            asyncio.ensure_future(
```
cancan101, line 233: does this code "work"? `observe_async_results` returns `None` and `asyncio.ensure_future` will raise when passed `None` (see: https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future).

PR #884, opened 2022-06-22
Fix: Add some Optional typing annotations

iyanging:


iyanging:
My local-venv mypy only got:
```
(.venv) ➜  ariadne git:(master) mypy ariadne --ignore-missing-imports
ariadne/contrib/tracing/opentracing.py:33: error: Item "None" of "Optional[Any]" has no attribute "close"
Found 1 error in 1 file (checked 35 source files)
```

But where does `ariadne/asgi.py:252: error: Argument 1 to "loads" has incompatible type "Union[UploadFile, str, None]"; expected "Union[str, bytes]"` come from ???

ariadne/contrib/tracing/opentracing.py
```diff
@@ -17,7 +17,7 @@
 
 class OpenTracingExtension(Extension):
     _arg_filter: Optional[ArgFilter]
-    _root_scope: Scope
+    _root_scope: Optional[Scope]
```
iyanging, line 20: see https://github.com/mirumee/ariadne/blob/63a5925ed5fbf9f9834cbd28f56e488873b56929/ariadne/contrib/tracing/opentracing.py#L26

rafalp:
Hi, can you rebase? Thank you!

rafalp:
I've implemented those manually. Thank you!

PR #905, opened 2022-08-05
Option to enable convert kwargs to snake case globally

enwawerueli:
A global option for the `convert_kwargs_to_snake_case` decorator would be nice to have. This is especially useful if this is the desired behavior for all resolvers

### DIFF
--- a/ariadne/executable_schema.py
+++ b/ariadne/executable_schema.py
@@ -6,6 +6,7 @@ from graphql import (
     build_ast_schema,
     parse,
 )
+from graphql.type import GraphQLObjectType, GraphQLInputObjectType
 
 from .enums import (
     EnumType,
@@ -14,12 +15,14 @@ from .enums import (
 )
 from .schema_visitor import SchemaDirectiveVisitor
 from .types import SchemaBindable
+from .utils import convert_camel_case_to_snake
 
 
 def make_executable_schema(
     type_defs: Union[str, List[str]],
     *bindables: Union[SchemaBindable, List[SchemaBindable]],
     directives: Dict[str, Type[SchemaDirectiveVisitor]] = None,
+    convert_args_to_snake_case: bool = False,
 ) -> GraphQLSchema:
     if isinstance(type_defs, list):
         type_defs = join_type_defs(type_defs)
@@ -39,6 +42,9 @@ def make_executable_schema(
     assert_valid_schema(schema)
     validate_schema_enum_values(schema)
     repair_default_enum_values(schema, flat_bindables)
+
+    if convert_args_to_snake_case:
+        set_outnames_to_snake_case(schema)
 
     return schema
 
@@ -65,3 +71,14 @@ def repair_default_enum_values(schema, bindables) -> None:
     for bindable in bindables:
         if isinstance(bindable, EnumType):
             bindable.bind_to_default_values(schema)
+
+
+def set_outnames_to_snake_case(schema) -> None:
+    for graphql_type in schema.type_map.values():
+        if isinstance(graphql_type, GraphQLInputObjectType):
+            for field_name, field in graphql_type.fields.items():
+                field.out_name = convert_camel_case_to_snake(field_name)
+        if isinstance(graphql_type, GraphQLObjectType):
+            for field in graphql_type.fields.values():
+                for arg_name, arg in field.args.items():
+                    arg.out_name = convert_camel_case_to_snake(arg_name)
--- a/ariadne/objects.py
+++ b/ariadne/objects.py
@@ -5,7 +5,6 @@ from graphql.type import GraphQLNamedType, GraphQLObjectType, GraphQLSchema
 from .resolvers import resolve_to
 from .types import Resolver, SchemaBindable
 
-
 class ObjectType(SchemaBindable):
     _resolvers: Dict[str, Resolver]
 
@@ -57,7 +56,7 @@ class ObjectType(SchemaBindable):
                 )
             if graphql_type.fields[field].resolve is None or replace_existing:
                 graphql_type.fields[field].resolve = resolver
-
+            
 
 class QueryType(ObjectType):
     """Convenience class for defining Query type"""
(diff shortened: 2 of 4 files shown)


rafalp:
Hi! There's better way to enable this behavior than through decorator. Every input field and argument has `out_name` attribute that can be set to custom value. Eg. if you set `out_name` on field `name` to `bob`, `name: "test"` in GraphQL query will be passed to `bob` kwarg/dict key. This is approach that Ariadne GraphQL Modules uses:

- Input fields: https://github.com/mirumee/ariadne-graphql-modules/blob/main/ariadne_graphql_modules/input_type.py#L117
- Fields args: https://github.com/mirumee/ariadne-graphql-modules/blob/main/ariadne_graphql_modules/object_type.py#L123

We could have an utility function that iterates schema types and if those types are `GraphQLInputType` or `GraphQLObjectType` we could set out names on those types. This util would be then called by `make_executable_schema` to do it's magic.

enwawerueli:
Noted. I haven't tried out graphql-modules, so I'll have a closer look at that first

ariadne/objects.py
```diff
@@ -57,7 +57,7 @@ def bind_resolvers_to_graphql_type(self, graphql_type, replace_existing=True):
                 )
             if graphql_type.fields[field].resolve is None or replace_existing:
                 graphql_type.fields[field].resolve = resolver
-
+
```
rafalp, line 60: Format code with black.

ariadne/executable_schema.py
```diff
@@ -40,6 +43,9 @@ def make_executable_schema(
     validate_schema_enum_values(schema)
     repair_default_enum_values(schema, flat_bindables)
 
+    if kwargs_to_snake_case:
```
rafalp, line 46: I'm curious if this should be an option or just a utility that you can run against your schema separately. But I guess the latter would require us to create a utility that copies whole schema so we don't mutate it.

ariadne/executable_schema.py
```diff
@@ -14,12 +15,14 @@
...
 from .schema_visitor import SchemaDirectiveVisitor
 from .types import SchemaBindable
+from .utils import convert_camel_case_to_snake
 
 
 def make_executable_schema(
     type_defs: Union[str, List[str]],
     *bindables: Union[SchemaBindable, List[SchemaBindable]],
     directives: Dict[str, Type[SchemaDirectiveVisitor]] = None,
+    kwargs_to_snake_case: bool = False
```
rafalp, line 25: How about `convert_args_to_snake_case`?

ariadne/executable_schema.py
```diff
@@ -65,3 +71,15 @@ def repair_default_enum_values(schema, bindables) -> None:
...
         if isinstance(bindable, EnumType):
             bindable.bind_to_default_values(schema)
+
+
+def set_outname_to_snake_case(schema) -> None:
+    for graphql_type in schema.type_map.values():
+        if not isinstance(graphql_type, (GraphQLObjectType, GraphQLInputObjectType)):
+            continue
+        for field_name, field in graphql_type.fields.items():
+            field.out_name = convert_camel_case_to_snake(field_name)
```
rafalp, line 81: We should do `if isinstance(graphql_type, GraphQLInputObjectType)` and then set field.out_name, then `if isinstance(graphql_type, GraphQLObjectType)` do that for args.

ariadne/executable_schema.py
```diff
@@ -65,3 +71,15 @@ def repair_default_enum_values(schema, bindables) -> None:
     for bindable in bindables:
         if isinstance(bindable, EnumType):
             bindable.bind_to_default_values(schema)
+
+
+def set_outname_to_snake_case(schema) -> None:
+    for graphql_type in schema.type_map.values():
+        if not isinstance(graphql_type, (GraphQLObjectType, GraphQLInputObjectType)):
```
rafalp, line 78: We should do `if isinstance(graphql_type, GraphQLInputObjectType)` and then set field.out_name, then `if isinstance(graphql_type, GraphQLObjectType)` do that for args.

rafalp (changes requested):
Looking okay, but could you please add tests?

ariadne/executable_schema.py
```diff
@@ -65,3 +71,15 @@ def repair_default_enum_values(schema, bindables) -> None:
     for bindable in bindables:
         if isinstance(bindable, EnumType):
             bindable.bind_to_default_values(schema)
+
+
+def set_outname_to_snake_case(schema) -> None:
```
rafalp, line 76:
```suggestion
def set_outnames_to_snake_cases(schema) -> None:
```

enwawerueli:
@rafalp I've incorporated the changes you suggested. Should be all good!

enwawerueli:
Also noted a name like `complexField_Int32` converts to `complex_field__int_32`, not sure if that's the intended behavior or it should be `complex_field_int_32`. Not a problem though

ariadne/executable_schema.py
```diff
@@ -65,3 +71,14 @@ def repair_default_enum_values(schema, bindables) -> None:
     for bindable in bindables:
         if isinstance(bindable, EnumType):
             bindable.bind_to_default_values(schema)
+
+
+def set_outnames_to_snake_case(schema) -> None:
```
rafalp, line 76:
```suggestion
def set_out_names_to_snake_case(schema) -> None:
```

ariadne/executable_schema.py
```diff
@@ -40,6 +43,9 @@ def make_executable_schema(
     validate_schema_enum_values(schema)
     repair_default_enum_values(schema, flat_bindables)
 
+    if convert_args_to_snake_case:
+        set_outnames_to_snake_case(schema)
```
rafalp, line 47:
```suggestion
        set_out_names_to_snake_case(schema)
```

rafalp:
Left some extra notes but those are tiny bits and nothing blocking - but code needs to be formatted with black to pass tests.

rafalp:
Thank you for contributing this. Can you say if approach from #1007 also works for you?

Thank you!

PR #910, opened 2022-08-10
Fix type signature for make_executable_schema

naartjie:
This PR changes `directives` to an `Optional`, since it can take a `None`.

It gets rid of this error message:

Screenshot 2022-08-10 at 13 57 19


naartjie:
I see it's already covered in #903

PR #936, opened 2022-09-30
fix #935 Remove duplicate validation error reported from the cost validator

### LINKED ISSUE #935: Cost validator throws duplicate error for an invalid input
### Discussed in https://github.com/mirumee/ariadne/discussions/934

<div type='discussions-op-text'>

<sup>Originally posted by **jayeshv** September 30, 2022</sup>
Consider the following app

```
from ariadne import QueryType, ScalarType, gql, make_executable_schema
from ariadne.validation import cost_directive, cost_validator

type_defs = gql("""
    scalar MyScalar

    type Query {
      test(myInput: MyScalar!): String
    }
""")

query = QueryType()
my_scalar = ScalarType("MyScalar")

@my_scalar.value_parser
def validate_myscalar(value):
    if value == "a":
        raise ValueError("Invalid myscalar")
    return value


@query.field("test")
def resolve_hello(_, info, **kw):
    return "hello"


schema = make_executable_schema([type_defs, cost_directive], query, my_scalar)
app = GraphQL(schema, debug=False, validation_rules=[cost_validator(maximum_cost=5)])

```
for the following query, it generates two error objects.

```
query test {
  test(myInput: "a") 
}
```
```
{
  "error": {
    "errors": [
      {
        "message": "Argument 'myInput' has invalid value \"a\".\n\nGraphQL request:2:17\n1 | query test {\n2 |   test(myInput: \"a\")\n  |                 ^\n3 | }"
      },
      {
        "message": "Expected value of type 'MyScalar!', found \"a\"; Invalid myscalar",
        "locations": [
          {
            "line": 2,
            "column": 17
          }
        ]
      }
    ]
  }
}
```
Isn't the first error created by the cost_validator a redundant one? This duplicate error happens only with the extra validator. 
ariadne version  0.16.1</div>

jayeshv:
I hope this explicit error can be removed safely as the cost validator has to report only its own errors.

### DIFF
--- a/ariadne/validation/query_cost.py
+++ b/ariadne/validation/query_cost.py
@@ -90,8 +90,7 @@ class CostValidator(ValidationRule):
                     field_args: Dict[str, Any] = get_argument_values(
                         field, child_node, self.variables
                     )
-                except Exception as e:
-                    report_error(self.context, e)
+                except Exception:
                     field_args = {}
                 if self.cost_map:
                     cost_map_args = (
--- a/tests/test_query_cost_validation.py
+++ b/tests/test_query_cost_validation.py
@@ -458,3 +458,18 @@ def test_child_field_cost_defined_in_directive_is_multiplied_by_values_from_lite
             extensions={"cost": {"requestedQueryCost": 20, "maximumAvailable": 3}},
         )
     ]
+
+
+def test_field_input_errors_are_not_reported_from_cost_validator(
+    schema_with_costs,
+):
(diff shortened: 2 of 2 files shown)


rafalp:
There are 6 places where error is raised, and out of those one is validation error, other is validation error for query cost map and remaining 4 are errors raised when query contents or AST is not what it should be. So there are four places that need taking care of, and just muting them doesn't feel right to me.

PR #937, opened 2022-10-01
install via pyproject.toml

singingwolfboy:
It's now possible to put all install metadata for a Python project into the `pyproject.toml` file, and [Flit](https://flit.pypa.io/en/stable/) makes it very straightforward to do so. This pull request deletes the `setup.py` file entirely, and migrates all the information there into the `pyproject.toml` file using Flit as the build backend.


rafalp:
Hello, thank you for this PR. I'm not familiar with this approach which makes me weary of merging it. Can you say if our current automation for releases will still work with this?

singingwolfboy:
The `pyproject.toml` file was first proposed in [PEP 518](https://peps.python.org/pep-0518/) back in 2016. `pip` added support for installing from `pyproject.toml` [in version 10, released in 2018](https://pip.pypa.io/en/stable/news/#v10-0-0). It is now considered to be the [standard build system interface](https://pip.pypa.io/en/stable/reference/build-system/), with `setup.py` officially declared as "legacy". [The `pip` documentation has more information about this format, as well.](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/)

All this is to say: yes, `pip install ariadne` will still work on older pip versions. Based on what I'm seeing in the [`publish.yml`](https://github.com/mirumee/ariadne/blob/master/.github/workflows/publish.yml) workflow, that should work just fine as well.

singingwolfboy:
@Rafalp is there anything else I can do to make you more comfortable with this change?

rafalp:
Hey @singingwolfboy, I think I'm okay with this change, thanks! I'll experiment with it when I am finished with my current work.

Would you be open to also making the PR to ariadne-graphql-modules repo?

rafalp:
I am sorry but can I ask you to rebase your PR, @singingwolfboy? I'll give it a priority then. Thanks!

singingwolfboy:
@rafalp Since making this pull request, I've discovered that [Hatch](https://hatch.pypa.io/latest/) is a nicer build backend than Flit; it does everything that Flit does, and a lot more! It's also supported by PyPA (the Python Packaging Authority), so I'm inclined to use it more in the future.

I've closed this pull request in favor of #963, which does the same thing using Hatch instead of Flit.

PR #960, opened 2022-11-01
fix 876: add support for @tag federation v1 directive

itamar82:
adds support for @tag direction in a federation v1 schema

### DIFF
--- a/tests/federation/test_schema.py
+++ b/tests/federation/test_schema.py
@@ -18,6 +18,31 @@ from ariadne.contrib.federation import (
 )
 
 
+def test_federation_one_schema_mark_type_tags():
+
+    type_defs = """
+        type Query
+        
+        type Product @tag(name: "test") {
+            upc: String!
+            name: String
+            price: Int @tag(name: "test2")
+        }
+    """
+    product = FederatedObjectType("Product")
+    schema = make_federated_schema(type_defs, product)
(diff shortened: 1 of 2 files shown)


tests/federation/test_schema.py
```diff
@@ -18,6 +18,31 @@
 )
 
 
+def test_federation_one_schema_mark_type_tags():
+
```
rafalp, line 22:
```suggestion
```

rafalp:
Is this all that it took to support `@tag` for federations? Is it only a directive used to put extra metadata on schema elements that has no related logic for supergraph members? I am asking this because I'm not up to date with Apollo's Federation these days. 🤔

itamar82:
> Is this all that it took to support `@tag` for federations? Is it only a directive used to put extra metadata on schema elements that has no related logic for supergraph members? I am asking this because I'm not up to date with Apollo's Federation these days. 🤔

AFAIK it is a directive used within the Apollo Gateway or when the Supergraph is composed so there shouldn't be any implementation necessary in the subgraph, just need to support it.

https://www.apollographql.com/docs/graphos/contracts/#2-fed1-enable-variant-support-for-tag

rafalp (approved):

rafalp:
Thank you!

PR #963, opened 2022-11-04
install via pyproject.toml (hatch)

singingwolfboy:
It's now possible to put all install metadata for a Python project into the `pyproject.toml` file, and [Hatch](https://hatch.pypa.io/latest/) makes it very straightforward to do so. This pull request deletes the `setup.py` file entirely, and migrates all the information there into the `pyproject.toml` file using Hatch as the build backend.

This also enables some new developer quality-of-life features. If you have Hatch installed on your system (such as via `brew install hatch`), you can run `hatch shell` to have Hatch automatically create a virtualenv for you, install the dependencies, and activate it. You can also run `hatch run test`, either inside or outside of that virtualenv, and Hatch will run the tests for you, setting up a development environment first if necessary. Of course, if you prefer not to use Hatch, you can still install this project with `pip`, exactly as you could do before.


rafalp (approved):

rafalp:
@singingwolfboy there's a problem: dependabot can no longer see Ariadne's dependencies and closed all PRs. Are you familiar with this?

rafalp:
Issue on dependabot: https://github.com/dependabot/dependabot-core/issues/3290

singingwolfboy:
> @singingwolfboy there's a problem: dependabot can no longer see Ariadne's dependencies and closed all PRs. Are you familiar with this?

I'm not, unfortunately. It seems like it should work; some of the most recent comments on https://github.com/dependabot/dependabot-core/issues/3290 are people saying that it's working for them, even with `pyproject.toml` and Hatch as the backend. Maybe there's some setting somewhere that needs to be changed? I wish I could help, but I'm not sure I can.

rafalp:
@singingwolfboy I've did a dive into this and it looks like what's happening here is by design.

Basically when you have `requirements.txt` file containing `dep=1.0.1`, dependabot knows that project uses `dep` in `1.0.1` version and it should step in when dep 1.0.2 is available. But when your dependency is `dep`, `dep>1` or `dep<3`, there's nothing for dependabot to do here because pip will install whatever version is available anyway.

So in a nutshell this PR de-facto unfroze our dependencies, rendering dependabot obsolete. I don't know yet if that's good thing.

rafalp:
I guess it's okay because Ariadne is a library and not a product, so final call on deps freeze should be in hands of maintainers who have Ariadne as dependency.

It would be good If we could run our tests whenever there's new release of one of dependencies on PIP, but I guess we'll have to do with weekly test run on mondays in the meantime.

PR #968, opened 2022-11-14
Add support for passing execution context class

yurihs:
Allows customizing the `ExecutionContext` by passing an argument to `graphql` and `graphql_sync`. If the argument is not provided, the default execution context class is used (as before).

I intend to use this with [graphql-sync-dataloaders](https://github.com/jkimbo/graphql-sync-dataloaders). AFAIK, it's the most convenient way to use dataloaders in a non-async Django project.

See also #817. Please let me know of any problems!

---

PS.: Type checks on master are failing with the latest mypy version (0.990):
```
ariadne/resolvers.py:64: error: Function "Callable[..., Any]" could always be true in boolean context  [truthy-function]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Added support for `@tag` directive used by Apollo Federation.
 - Updated `starlette` dependency in setup.py to `<1.0`.
 - Moved project configuration from `setup.py` to `pyproject.toml`.
+- Added `execution_context_class` option.
 
 
 ## 0.16.1 (2022-09-26)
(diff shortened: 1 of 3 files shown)


rafalp:
Can you add both async and sync dataloaders to dev dependencies and add a `test_dataloaders.py` to the tests suite with two tests:

- `def test_graphql_supports_async_dataloaders`
- `def test_graphql_sync_supports_sync_dataloaders`

Thanks!

yurihs:
> Woah, amazing! This is something we really wanted to do for 0.17! Thank you for opening PR!

Nice! Love to contribute :heart:

> Can you add both async and sync dataloaders to dev dependencies and add a `test_dataloaders.py` to the tests suite with two tests

Done! There is a bit of code duplication, but I think it's OK... Let me know.

PR #972, opened 2022-11-16
fixed websocket.close call on already closed websocket instance

pigletto:
"Unexpected ASGI message 'websocket.close', after sending 'websocket.close'." exception appears in the server logs.
This is caused by the handle_connection_init_timeout handler that is not checking if the WebSocket is already disconnected.


rafalp:
Thank you!

PR #1005, opened 2023-01-01
Make load-schema functions accept path-like objects

pylipp:
Otherwise, a call like `load_schema_from_path(dir)`, with `dir = pathlib.Path(...)`, is being reported by mypy as an error.

### DIFF
--- a/ariadne/load_schema.py
+++ b/ariadne/load_schema.py
@@ -7,16 +7,14 @@ from graphql.error import GraphQLSyntaxError
 from .exceptions import GraphQLFileSyntaxError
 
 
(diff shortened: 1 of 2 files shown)


rafalp (approved):

rafalp (changes requested):
`str(b"bytes")` produces `"b'bytes'"`, not "bytes"

rafalp:
We shouldn't support bytes, we can't assume what kind of encoding the string passed to us is.

rafalp (approved):

rafalp:
Note to self: merge this and add changelog item later.

rafalp:
Thank you!

PR #1040, opened 2023-02-21
Update typing for values in EnumType.__init__

### LINKED ISSUE #1039: Incorrect typing for values on EnumType.__init__()
Ariadne 0.18.0 has updated the `EnumType` class in a way that causes the `values` argument of `__init__()` to become typed, but the typing is incorrect:

```
    def __init__(
        self, name: str, values: Union[Dict[str, Any], enum.Enum, enum.IntEnum]
    ) -> None:
```

should be:

```
    def __init__(
        self, name: str, values: Union[Dict[str, Any], Type[enum.Enum], Type[enum.IntEnum]]
    ) -> None:
```

markedwards:
Fixes #1039

### DIFF
--- a/ariadne/enums.py
+++ b/ariadne/enums.py
@@ -7,6 +7,7 @@ from typing import (
     Mapping,
     Optional,
     Tuple,
+    Type,
     TypeVar,
     Union,
     cast,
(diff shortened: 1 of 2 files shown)


rafalp:
Can you also add `enums.py` test to `tests_mypy` directory?

markedwards:
> Can you also add `enums.py` test to `tests_mypy` directory?

Missed that.  Amended to the commit.

rafalp (approved):

rafalp:
Thank you! I'll look into releasing those tomorrow as 0.18.1

rafalp (approved):

PR #1049, opened 2023-03-15
Fix federation support for directives without surrounding whitespace

danring:
As written, this regex requires two delimiters before a `directive @custom on FIELD` definition. In some cases (for example, when the preceding line is a comment), this causes `purge_schema_directives` to remove only the `@custom` part, leaving `directive on FIELD` which causes downstream parsing errors.

The solution is to require the second delimiter only if the description is also present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added support for passing `Enum` types directly to `make_executable_schema`
 - Added `convert_names_case` option to `make_federated_schema`.
 - Added support for the `@interfaceObject` directive in Apollo Federation.
+- Fixed federation support for directives without surrounding whitespace.
 
 
 ## 0.18.1 (2023-02-22)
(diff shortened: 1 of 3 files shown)


rafalp:
I am wondering if this isn't the time to stop using the regexes spagetti and build clean solution on top of `parse` and GraphQL AST?

rafalp:
Linters are failing for this change.

danring:
I split the long line, so linters are passing now.

rafalp:
Can you add unit test for the issue you are fixing?

danring:
There is already a test added [here](https://github.com/mirumee/ariadne/pull/1049/files#diff-38572c5999e8f660100ce4742b3947b3fc0d015487579cadfe40a0835cba33d9R124) for the specific case this fixes -- do you want an additional test somewhere else?

rafalp:
I'm blind, sorry 😅

Final thing would be updating changelog to include a note about the issue being fixed.

danring:
Should I also bump to 18.2? Or just add a new "since last release" section in the changelog?

rafalp:
@danring there's already "next release" section. If you don't see it in your files you'll have to rebase.

danring:
Rebased and added an entry to CHANGELOG

rafalp:
Please rebase

danring:
Done

rafalp (approved):

PR #1056, opened 2023-03-21
Add convert names case to make federated schema

a-rhote:
Attempt 2, with more imports this time :D

### DIFF
--- a/ariadne/contrib/federation/schema.py
+++ b/ariadne/contrib/federation/schema.py
@@ -10,12 +10,12 @@ from graphql.type import (
     GraphQLUnionType,
 )
 
-from ... import SchemaNameConverter
 from ...executable_schema import (
     SchemaBindables,
     make_executable_schema,
     join_type_defs,
 )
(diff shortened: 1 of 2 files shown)


rafalp (approved):

rafalp:
Okay, now those are running

ariadne/contrib/federation/schema.py
```diff
@@ -10,6 +10,7 @@
     GraphQLUnionType,
 )
 
+from ... import SchemaNameConverter
```
rafalp, line 13: Make this import use `schema_names` module and reorder imports.

rafalp (changes requested):

ariadne/contrib/federation/schema.py
a-rhote: Can do

a-rhote: Should be updated now

rafalp (approved):

rafalp:
Thank you!

PR #1057, opened 2023-03-21
Add `InputType`

### LINKED ISSUE #333: Bind Python Types to GraphQL Input Types [How To]
Perhaps it's just unclear in the docs, but I don't see any way to automatically convert from the input dictionaries given by Ariadne, to my resolvers, into custom Python types (i.e. dataclasses)

Is this supported? I found the docs on bindables, but from just that I can't tell if that's what I want. 

Example

```python
@dataclass
class Person:
    id: int
    name: str
    phone_number: str


@mutation.field('updatePerson')
def update_person(_, info, person: Person):
    # Currently what comes in is a dict. 
    # Is there a way to automatically map this to Person?
    do_update(
        person.id, 
        name=person.name, 
        phone=person.phone_number
    )
```

rafalp:
Adds `InputType` which enables setting custom Python representations for GraphQL inputs, eg:

```python
@dataclass
class InputDataclass:
    id: str
    message: str
    year_of_birth: int


input_type = InputType(
    "ExampleInput",
    lambda data: InputDataclass(**data),
    {"yearOfBirth": "year_of_birth"},
)
```

Fixes #333

# TODO:

- [ ] Docs

### DIFF
--- a/ariadne/inputs.py
+++ b/ariadne/inputs.py
@@ -178,7 +178,7 @@ class InputType(SchemaBindable):
         graphql_type = cast(GraphQLInputObjectType, graphql_type)
 
         if self._out_type:
-            graphql_type.out_type = self._out_type  # pylint: disable
+            graphql_type.out_type = self._out_type  # type: ignore
 
         if self._out_names:
             for graphql_name, python_name in self._out_names.items():
--- a/ariadne/inputs.py
+++ b/ariadne/inputs.py
@@ -1,7 +1,7 @@
 from typing import Dict, Optional, cast
 
 from graphql import GraphQLInputObjectType, GraphQLSchema
-from graphql.type.definition import GraphQLInputFieldOutType
+from graphql.type.definition import GraphQLInputFieldOutType, GraphQLNamedType
 
 from .types import SchemaBindable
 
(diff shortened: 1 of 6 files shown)


ariadne/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+        InputType("ExampleInput", lambda data: ExampleInput(**data)),
+    )
+    ```
+
+    # Example input with its fields mapped to custom dict keys
+
+    Following code creates a GraphQL schema with object type named `Query`
+    with single field which has an argument of an input type. It then uses
+    the `InputType` to set custom "out names" values, mapping GraphQL
+    `shortMessage` to `short_message` key in Python dict:
```
mat-sop, line 59:
```suggestion
    `shortMessage` to `message` key in Python dict:
```
In example code `shortMessage` is mapped to `message` not `short_message`

ariadne/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+
+    ```python
+    from ariadne import InputType, QueryType, make_executable_schema
+
+    query_type = QueryType()
+
+    @query_type.field("repr")
+    def resolve_repr(*_, input: dict):
+        # Dict will have `id` and `short_message` keys
```
mat-sop, line 68:
```suggestion
        # Dict will have `id` and `message` keys
```

mat-sop (approved):

PR #1060, opened 2023-03-22
feat: add support for @interfaceObject directive

moonflare:
### Context
Apollo release a new version of Federation, version 2.3. Part of this release is the support for the `@interfaceObject` directive, which will allow users to extend/reference Entities from other subgraphs that are interfaces. You can read more about `@interfaceObject` directive here: https://github.com/mirumee/ariadne/issues/1032

This PR is an initial step into adding support for newer spec versions (`v2.1`, `v2.2` and `v2.3`), but the changes are very minimal so we unlock the work we have in progress regarding `Interface Entities`.

### Description

* This PR only adds definitions for all the newer spec versions (`v2.1`, `v2.2` and `v2.3`), but it doesn't add functionality for some of the features part of these new spec versions (for example `@composeDirective`), so it will unblock our migration from `Federation v1` to `Federation v2.3`

### Next steps

The next iterations for this work will be to add support for some more features part of the new specs. I have some of the work in progress, but it's not near being done.

Some of the features that will be covered in following PRs:
* Support for `@composeDirective` (Currently, we [remove](https://github.com/mirumee/ariadne/blob/master/ariadne/contrib/federation/schema.py#L86) all the custom directives, we want to keep directives defined with `@composeDirective`)
* Validation for directives like `@composeDirective` and `@interfaceObject`, to ensure we are using them with the right spec version
* Support for namespaces
```GraphQL
extend schema @link(url: "https://specs.apollo.dev/federation/v2.3", as: "fed2", import: ["@inaccessible", "@tag"])
            
type Product @fed2__key(fields: "upc") {
    upc: String!
  }
  
  type Query {
    product: Product
  }
```
* Renamed imports
```GraphQL
extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: [{ name: "@key", as: "@myKey" }, { name: "@shareable" }, "@provides","@inaccessible"])

type Product @myKey(fields: "id") {
    id: ID!
    sku: String @shareable
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added `InputType` for setting Python representations of GraphQL Input types
 - Added support for passing `Enum` types directly to `make_executable_schema`
 - Added `convert_names_case` option to `make_federated_schema`.
+- Added support for the `@interfaceObject` directive in Apollo Federation.
 
 
 ## 0.18.1 (2023-02-22)
--- a/ariadne/contrib/federation/definitions/fed1_0.graphql
+++ b/ariadne/contrib/federation/definitions/fed1_0.graphql
@@ -1,0 +1,6 @@
+directive @key(fields: _FieldSet!) repeatable on OBJECT | INTERFACE
+directive @requires(fields: _FieldSet!) on FIELD_DEFINITION
+directive @provides(fields: _FieldSet!) on FIELD_DEFINITION
+directive @external on FIELD_DEFINITION
+directive @extends on OBJECT | INTERFACE
+scalar _FieldSet
--- a/ariadne/contrib/federation/schema.py
+++ b/ariadne/contrib/federation/schema.py
@@ -1,4 +1,5 @@
 import re
+import os
 from typing import Dict, List, Optional, Type, Union, cast
 
 from graphql import extend_schema, parse
@@ -17,6 +18,7 @@ from ...executable_schema import (
 )
 from ...schema_names import SchemaNameConverter
 from ...schema_visitor import SchemaDirectiveVisitor
+from ...load_schema import load_schema_from_path
 from .utils import get_entity_types, purge_schema_directives, resolve_entities
 
 base_federation_service_type_defs = """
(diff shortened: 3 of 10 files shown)


ariadne/contrib/federation/schema.py
```diff
@@ -26,26 +26,48 @@
...
-    directive @external on FIELD_DEFINITION
-	directive @requires(fields: String!) on FIELD_DEFINITION
-	directive @provides(fields: String!) on FIELD_DEFINITION
-	directive @extends on OBJECT | INTERFACE
+    directive @extends on OBJECT | INTERFACE
+    directive @external on OBJECT | FIELD_DEFINITION
+    directive @requires(fields: FieldSet!) on FIELD_DEFINITION
+    directive @provides(fields: FieldSet!) on FIELD_DEFINITION
+    directive @tag(name: String!) repeatable on FIELD_DEFINITION | INTERFACE | OBJECT | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+    scalar FieldSet
```
dariuszkuc, line 34: Federation v1 actually uses `_FieldSet` scalar, it was renamed to `FieldSet` in v2 (guessing to avoid weird naming when namespacing it which would result in two underscores, i.e. `federation__FieldSet` vs `federation_FieldSet`)

ariadne/contrib/federation/schema.py
```diff
@@ -26,26 +26,48 @@
         _service: _Service!
     }}
 
-    directive @external on FIELD_DEFINITION
-	directive @requires(fields: String!) on FIELD_DEFINITION
-	directive @provides(fields: String!) on FIELD_DEFINITION
-	directive @extends on OBJECT | INTERFACE
+    directive @extends on OBJECT | INTERFACE
+    directive @external on OBJECT | FIELD_DEFINITION
```
dariuszkuc, line 30: `@external` was only applicable on `FIELD_DEFINITION` in v1, v2 added support for `OBJECT`

dariuszkuc (changes requested):

ariadne/contrib/federation/definitions/fed1_0.graphql
```diff
@@ -0,0 +1,17 @@
+directive @key(fields: _FieldSet!) repeatable on OBJECT | INTERFACE
+directive @requires(fields: _FieldSet!) on FIELD_DEFINITION
+directive @provides(fields: _FieldSet!) on FIELD_DEFINITION
+directive @external on FIELD_DEFINITION
+directive @extends on OBJECT | INTERFACE
+directive @tag(name: String!) repeatable on
```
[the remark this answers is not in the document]
moonflare: Makes sense. I've updated the tests to still keep that as reference, but we will need to document this somehow I guess.

rafalp:
Can you update the PR to include changelog entry for your changes?

moonflare:
> Can you update the PR to include changelog entry for your changes?

Thanks @rafalp. Added.
Please let me know if there's anything else needed 🙏🏻

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Next (UNRELEASED)
 
+- Added support for the `@interfaceObject` directive
```
rafalp, line 5:
```suggestion
- Added support for the `@interfaceObject` directive in Apollo Federations
```

rafalp (changes requested):

rafalp (approved):

Issue #859, opened 2022-05-03
Make it possible to exempt some resolvers from tracing


vlaci:
The default behavior of only tracing custom resolvers is great 95% of the time, however there is a case for not to trace certain custom resolvers as well.

We have a few simple custom resolvers that can be called up to a couple thousand times per request where the tracing overhead becomes significant. Also, in our case they don't do meaningful work that would warrant tracing them.

Right now I see two ways to implement a workaround in our code-base without touching ariadne itself:

1. Monkey-patch `ariadne.contrib.tracing.should_trace` function
2. Add `_ariadne_alias_resolver` attribute to our own resolvers, disguising them as alias resolvers which are already exempt fro tracing.
3. Extend the tracing extensions in our code to do the filtering. 

The first two approaches are brittle and can break upon any update to ariadne. We can go forward using the third approach but we think it'd make sense to add this functionality in ariadne itself.

I'd like to contribute support for this functionality in ariadne if you accept my reasoning for the need of it. My plan is to add a `filter` callable to the tracing extensions that can look-up the field in a block-list or perform any other custom logic.
An even less intrusive change could also be just adding a `should_trace` method in the tracing extensions that can be overridden in a subclass, instead of using a non-member function.

rafalp:
Construction option (like we are already doing with `args_filter`) is the way to go here IMHO. This option should default to `should_trace`.

vlaci:
I've taken a quick look and it is easy to do this way but the `should_trace` function having a `trace_default_resolver` bool argument makes it a bit smelly to me. One part of the tracing is configured via a callable, the other via a separate flag seems weird. It is easy to make `should_trace` be more restrictive via wrapping it to a separate function but it is hard to make it more permissive without duplicating its contents.

rafalp:
Yeah, we should drop `trace_default_resolver` at same time, but it would be good to have both at same time for deprecation period.

For now we could have both options but raise an error if custom tracing filter is set together with `trace_default_resolver`.

rafalp:
Faster extensions landed in Ariadne 0.20, but as [I've mentioned on Ariadne's blog](https://ariadnegraphql.org/blog/2023/07/10/graphql-gathering#performance-metrics), there's still utility in having an option to remove fields from tracing extensions for removing the especially noisy resolvers from APM.

PR #1066, opened 2023-03-30
Custom should_trace predicate in opentracing extensions

studzien:
Hey :wave:

This is my take on #859.

At Whatnot, we are facing a similar problem to the one described, where we have a few resolvers implemented on some custom types.
The penalty for tracing every single resolver, especially on a long list, is very significant, and we're hoping to reduce response times by not generating spans for some of them.

### DIFF
--- a/ariadne/contrib/tracing/opentracing.py
+++ b/ariadne/contrib/tracing/opentracing.py
@@ -20,15 +20,23 @@ except ImportError:
 
 
 ArgFilter = Callable[[Dict[str, Any], GraphQLResolveInfo], Dict[str, Any]]
+ShouldTrace = Callable[[GraphQLResolveInfo], bool]
 
 
 class OpenTracingExtension(Extension):
     _arg_filter: Optional[ArgFilter]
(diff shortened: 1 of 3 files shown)


rafalp:
Thanks for opening a PR. I'll flag this with decision needed because I would like to explore some ideas for faster middlewares first before committing to this option.

rafalp:
> The penalty for tracing every single resolver, especially on a long list, is very significant, and we're hoping to reduce response times by not generating spans for some of them.

Most of penalty for tracing comes from asynchronous middleware making all field resolutions in graphql query executor asynchronous. `async`/`await` overhead is non-negligible when there are thousands of resolver calls.

But you may have some noise resolvers you may want to exclude from tracing anyway, so we could support that. However we are deprecating opentracing because the standard was declared deprecated and superseded by opentelemetry, for which we will be adding an extension in next Ariadne release.

rafalp:
Closing this for lack of activity. @studzien if you want to update this PR to work for both OpenTracing and OpenTelemetry, please let me know here and I'll reopen this 😄

PR #1075, opened 2023-04-20
Run benchmarks periodically and rotate their results

rafalp:
This PR adds new workflow fired every monday, that benchmarks Ariadne against every Python version in the matrix. It stores benchmark results and commits them to master.

There's also small `rotate_results.py` script that will keep number of benchmark results files per version under control (no more than 30)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,43 @@
+name: Benchmarks
+
+on:
+  push:
+    branches:
+      - master
+  schedule:
+    - cron:  "0 7 * * 1,3"
```
mat-sop, line 8: In description, you wrote ` fired every monday,`
```suggestion
    - cron:  "0 7 * * 1"
```

mat-sop (approved):

PR #1077, opened 2023-04-22
update type-hints

AjmalPonneth:
`|` only support python 3.10 and above versions.

### DIFF
--- a/generate_reference.py
+++ b/generate_reference.py
@@ -5,6 +5,7 @@ import re
 from dataclasses import dataclass
 from importlib import import_module
 from textwrap import dedent
+from typing import Union
 
 import ariadne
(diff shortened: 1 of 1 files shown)


rafalp:
Reference generator is a devtool that is supposed to be ran by developer on their machine. How did you run into this?

AjmalPonneth:
@rafalp I am using python 3.9 on my machine. So I got an error in Pycharm for this `reference_generator.py` file and I just checked the issue also Ariadne says python version 3.7 and higher in docs. So I thought to make a PR by changing to Union types.

rafalp (approved):

rafalp:
Oh, okay. Thanks!

PR #1096, opened 2023-06-06
Remove Python 3.7 support

rafalp:
Python 3.7 support ends on 27th of June, in 3 weeks. Flask already abandoned support for it in latest release, I think its safe for us to do it too.


mat-sop (commented):
Some tests have also logic for python 3.7 [here](https://github.com/mirumee/ariadne/blob/master/tests/test_dataloaders.py#L10) and [here](https://github.com/mirumee/ariadne/blob/master/tests/wsgi/test_configuration.py#L21)

PR #1123, opened 2023-10-05
remove max version limit for graphql-core

tonyrewin:
then gql will support more async transports


rafalp:
I don't like this change on a basis that people installing Ariadne will have to keep up themselves with it's compatibility with future GraphQL Core releases. And when those are incompatible, we will be the first ones to take heat for it, as we've been in the past.

rafalp:
Which async transport do you want to use that's currently not supported by Ariadne?

tonyrewin:
> Which async transport do you want to use that's currently not supported by Ariadne?
> 
> I understand you want to use GraphQL Core 3.3 alpha - other ways exist to do this like having custom version of Ariadne on GitHub and installing that one instead of `ariadne` from Pypi.

i wanted to use httpx for the latest only version of gql

rafalp:
Sorry for late response but this is **not** going to happen. If you need Ariadne with GraphQL Core 3.3, please make a fork. Unpinning this dependency from Ariadne is a recipe for disaster.

PR #1158, opened 2024-02-14
Add support for federation versions up to 2.6

nogates:
Added support so we can parse fed 2.4, 2.5 and 2.6. schemas... 

Though I haven't added any extra code to mange those new directives or anything like that .

Also, refactored a bit the logic to fallback to the latest avaiable version when using `links` directive, which I think it's better comparing with defaulting to Federation v1


nogates:
@rafalp Could you have a look when you got a sec? 🙏

ariadne/contrib/federation/schema.py
```diff
@@ -76,25 +76,27 @@ def make_federated_schema(
...
-    elif link and link.group(1) == "https://specs.apollo.dev/federation/v2.3":
-        definitions = load_schema_from_path(
-            os.path.join(dirname, "./definitions/fed2_3.graphql")
+    definitions_folder = os.path.join(dirname, "definitions")
+    definition_files = os.listdir(definitions_folder)
+    definition_files.sort()
+    last_version_file = definition_files[-1]
+
+    if link:
+        # extract version
```
rafalp, line 85:
```suggestion
```

ariadne/contrib/federation/schema.py
```diff
@@ -76,25 +76,27 @@ def make_federated_schema(
...
+        fed_version = link.group(1).split("/")[-1]
```
rafalp, line 86: Doesn't this change mean that `https://reddit.com/r/something/v2.1` will also pass as Federation v2.1?

nogates: oh yeah! I should validate the link url and that it belongs to the fed spec. good catch! 🙏

nogates:
ready for another review @rafalp. Thanks!

rafalp (approved):

rafalp:
Thanks!

nogates:
Thanks @rafalp ! May I ask, when are you planning in creating a release? Happy to create a PR to fill in the CHANGELOG if that helps! Thanks!

rafalp:
When its ready™ Goal is to have query execution over GET in next Ariadne release. When that's merged, I'll cut out a release.

nogates:
Awesome, thank you!

rafalp:
It was already part of 0.23 beta 1, but I'll ship final thing today.

PR #1170, opened 2024-02-28
Add hack for updating final result with data from root value

### LINKED ISSUE #1167: Return data and errors from root resolver callable
Related to: https://github.com/mirumee/ariadne-graphql-proxy/issues/45

~~Root resolver could return a tuple of `data` and `error` or `GraphQLResult`. Errors would then be merged with final result.~~

New idea: a dedicated class (eg. `RootValueResult`) that will implement a `create_result` method. This

rafalp:
Fixes #1167

Skips changelog because its kind of private-ish API.

### DIFF
--- a/ariadne/graphql.py
+++ b/ariadne/graphql.py
@@ -144,10 +144,10 @@ async def graphql(
     `execution_context_class`: `ExecutionContext` class to use by query
     executor.
 
-    result_update: Optional[BaseProxyRootValue] = None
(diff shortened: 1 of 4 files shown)


rafalp:
Tests are passing but I am not entirely happy with the `GraphQLResultUpdate` name and location in `types.py`. I will sleep over with it and do a final update tomorrow.

PR #1227, opened 2025-02-04
feat!: add a default global ID encoder, add a id_field on the query type

pkucmus:
BREAKING CHANGE: the global ID function takes only a str with the GID, the query type uses the `id_field` arg to determine which kwarg to pass to the decoder

One would be able to do:

```python
query = RelayQueryType(
    id_field="bid",
)
```
Instead of this:

```python
def decode_global_id(kwargs) -> GlobalIDTuple:
    return GlobalIDTuple(*b64decode(kwargs["bid"]).decode().split(":"))


query = RelayQueryType(
    global_id_decoder=decode_global_id,
)
```

With the following: https://github.com/mirumee/ariadne/blob/2c8f064c4538bd26e709ed8da665c2236fcbb906/ariadne/contrib/relay/objects.py#L122

This is to make the decoder a bit dummer as to take the responsibility of pulling the GID from the right kwarg (or query param, if you will) from the decoder and leave it with the single responsibility.

### DIFF
--- a/ariadne/contrib/relay/utils.py
+++ b/ariadne/contrib/relay/utils.py
@@ -1,0 +1,13 @@
+from base64 import b64decode, b64encode
+
+from ariadne.contrib.relay.types import (
+    GlobalIDTuple,
+)
+
+
+def decode_global_id(gid: str) -> GlobalIDTuple:
+    return GlobalIDTuple(*b64decode(gid).decode().split(":"))
+
+
+def encode_global_id(type_name: str, _id: str) -> str:
+    return b64encode(f"{type_name}:{_id}".encode()).decode()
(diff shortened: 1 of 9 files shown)


ariadne/contrib/relay/utils.py
```diff
@@ -0,0 +1,13 @@
...
+from ariadne.contrib.relay.types import (
+    GlobalIDTuple,
+)
+
+
+def decode_global_id(gid: str) -> GlobalIDTuple:
+    return GlobalIDTuple(*b64decode(gid).decode().split(":"))
+
+
+def encode_global_id(type_name: str, _id: str) -> str:
```
KarolJagodzinski, line 12: Can we add an unittest for that?

KarolJagodzinski (approved):

PR #1228, opened 2025-02-06
use graphql core extensions field to store contrib resolvers

reallistic:
This pull request uses the `object_type.extensions` argument to store the resolvers used by Relay and Federation.
The `extensions` argument gets copied when `graphql.extend_schema` is used, ensuring the resolvers are kept around when schemas evolve dynamically. The added attributes done via `setattr` are not copied. The federation module worked around this by providing a `extend_federated_schema` function which copied those extra arguments. This function is no longer needed and now throws a deprecation warning.

### DIFF
--- a/ariadne/contrib/federation/interfaces.py
+++ b/ariadne/contrib/federation/interfaces.py
@@ -1,10 +1,14 @@
+from typing import cast
 from typing import Optional
 
+from graphql import GraphQLNamedType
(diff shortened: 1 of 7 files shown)


pkucmus (approved):
LGTM, thanks!

reallistic:
Thanks for approving @pkucmus . I do not have write access so please merge this at your convenience.

PR #1237, opened 2025-03-01
Fixed using GraphiQLPluginExplorer

iafilin:


kevinvalk:
Note to self: always important to check open PRs! Anyways, you removed a bit too much here. You can still pass on the query and onEdit parameters.

```jinja2
        {% if enable_explorer_plugin %}
        var explorerPlugin = GraphiQLPluginExplorer.explorerPlugin({
          query: query,
          onEdit: setQuery,
        });
        {% endif %}
```

That said, it seems that query and setQuery is not needed anymore so the whole `AriadneGraphiQL` function can be simplified to
```jinja2
      function AriadneGraphiQL() {
        {% if enable_explorer_plugin %}
        var explorerPlugin = GraphiQLPluginExplorer.explorerPlugin();
        {% endif %}
        return React.createElement(GraphiQL, {
          fetcher: fetcher,
          defaultEditorToolsVisibility: true,
          {% if enable_explorer_plugin %}
          plugins: [explorerPlugin],
          {% endif %}
        });
      }
```

@iafilin would you mind adjusting this PR? 

P.S. You also need to fix the test snapshots, do you need any help for that?

DamianCzajkowski:
Hi @iafilin @kevinvalk, are you still interested in completing this PR, or would you like some assistance with getting it merged?

PR #1334, opened 2026-07-30
fix: reject start before connection_init in legacy graphql-ws handler AR-13

Minister944:
GraphQLWSHandler dispatched `start` messages straight to subscription
execution without checking whether `connection_init` had completed,
so on_connect never ran for a client that sent `start` first. Track
connection_acknowledged state and close with code 4401 when `start`
arrives before the connection has been acknowledged, matching the
existing guard in GraphQLTransportWSHandler.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced the WebSocket handshake before allowing subscriptions to start.
  * Connections that send subscription requests before acknowledgment now close with code `4401`.
  * Prevented connection callbacks and subscription processing when initialization is unsuccessful or incomplete.
  * Fixed resolver-binding errors for scalar and list-based fields.
* **Tests**
  * Added coverage for invalid handshake order and confirmed connection callbacks are not invoked.

### DIFF
--- a/ariadne/asgi/handlers/graphql_ws.py
+++ b/ariadne/asgi/handlers/graphql_ws.py
@@ -114,9 +114,26 @@ class GraphQLWSHandler(GraphQLWebsocketHandlerBase):
                 websocket.application_state,
             ):
                 message = await websocket.receive_json()
-                connection_acknowledged = await self.handle_websocket_message(
-                    websocket, message, operations, connection_acknowledged
-                )
+                message_type = message.get("type")
+
+                if (
+                    message_type == GraphQLWSHandler.GQL_START
+                    and not connection_acknowledged
+                ):
+                    # 4401: Unauthorized. `start` received before a successful
+                    # `connection_init`/`connection_ack` handshake -- refuse it
+                    # rather than letting it reach subscription execution and
+                    # skip the `on_connect` hook entirely.
+                    await websocket.close(code=4401)
+                    break
+
+                await self.handle_websocket_message(websocket, message, operations)
+
+                if message_type == GraphQLWSHandler.GQL_CONNECTION_INIT:
+                    connection_acknowledged = WebSocketState.DISCONNECTED not in (
+                        websocket.client_state,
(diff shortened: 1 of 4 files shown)


tests/asgi/test_websockets_graphql_ws.py
```diff
@@ -789,6 +790,43 @@ def test_websocket_connection_can_be_kept_alive(
...
+            }
+        )
+
+        with pytest.raises(WebSocketDisconnect) as exc_info:
+            ws.receive_json()
+
+        assert exc_info.value.code == 4401
+
+
+def test_start_before_connection_init_does_not_call_on_connect(client):
```
Kwaidan00: Are you sure we need it as a separate test case? Can't this additional assertion included in the test case above?

ariadne/asgi/handlers/graphql_ws.py
```diff
@@ -106,14 +106,34 @@ async def handle_websocket(self, websocket: WebSocket):
...
+                    # `connection_init`/`connection_ack` handshake -- refuse it
+                    # rather than letting it reach subscription execution and
+                    # skip the `on_connect` hook entirely.
+                    await websocket.close(code=4401)
+                    break
+
                 await self.handle_websocket_message(websocket, message, operations)
+
+                if message_type == GraphQLWSHandler.GQL_CONNECTION_INIT:
+                    connection_acknowledged = WebSocketState.DISCONNECTED not in (
```
Kwaidan00, line 133: Why do you need to check the websocket state here? Why can't you just `connection_acknowledged = True`?

Minister944: If on_connect rejects the connection (raises), handle_websocket_connection_init_message sends connection_error and closes the socket instead of sending connection_ack. Setting connection_acknowledged = True unconditionally here would mark the connection as acknowledged even in that rejection case, which defeats the point of the start-before-ack check above.

DamianCzajkowski (approved):

PR #1335, opened 2026-08-21
fix: count list-valued `@cost` multipliers by their length

jaideeppyne:
## What

The query-cost validator (a DoS-protection control) ignores **list-valued** `@cost` multipliers, so a field's cost collapses to just its `complexity` no matter how large the list argument is — letting a large query slip past `maximum_cost`.

```graphql
# @cost(complexity: 1, multipliers: ["ids"]) — should cost len(ids)
{ things(ids: [ /* 10000 ids */ ]) }   # actually costs 1
```

## Why

In `CostValidator.get_multipliers_from_string` the length step is **dead code**:

```python
try:
    if val is not None and not isinstance(val, dict):
        parsed_vals.append(int(val))        # int([...]) raises TypeError...
except (ValueError, TypeError):
    pass                                    # ...which is swallowed here
parsed_vals = [
    len(multiplier) if isinstance(multiplier, list | tuple) else multiplier  # never sees lists
    for multiplier in parsed_vals
]
```

For a list argument, `int(val)` raises `TypeError`, which the `except` swallows, so the list is dropped **before** the `len()` comprehension runs. `parsed_vals` therefore only ever contains ints, and the intended `len(list)` contribution is lost. This affects both the `@cost` directive path and the `cost_map` path (both call this function).

## How

Handle list/tuple arguments by appending their length *before* the int coercion, and drop the now-redundant post-loop comprehension:

```diff
-            try:
-                if val is not None and not isinstance(val, dict):
-                    parsed_vals.append(int(val))
-            except (ValueError, TypeError):
-                pass
-        parsed_vals = [
-            len(multiplier) if isinstance(multiplier, list | tuple) else multiplier
-            for multiplier in parsed_vals
-        ]
-        return [m for m in parsed_vals if m > 0]
+            if isinstance(val, list | tuple):
+                parsed_vals.append(len(val))
+            elif val is not None and not isinstance(val, dict):
+                try:
+                    parsed_vals.append(int(val))
+                except (ValueError, TypeError):
+                    pass
+        return [m for m in parsed_vals if m > 0]
```

## Tests

Added `test_cost_directive_multiplier_counts_list_argument_length` and `test_cost_map_multiplier_counts_list_argument_length`: a `things(ids: [ID!]) @cost(complexity: 1, multipliers: ["ids"])` field with a 5-element list is rejected at `maximum_cost=4` and allowed at `maximum_cost=5`. Both **fail on `main`** (the query is accepted, cost computed as 1) and pass with this change; the full `test_query_cost_validation.py` suite stays green (34 passed).

---

*Disclosure: this change was prepared with AI assistance and reviewed/verified by me before submission.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Query cost calculations now use the length of list and tuple arguments as multipliers.
  * Non-numeric multiplier arguments no longer add unintended cost; fields are charged only their base complexity.
  * Cost calculations now handle list-valued arguments consistently across directive and configuration-based costing.

* **Documentation**
  * Updated query cost validation guidance with list-multiplier examples and upgrade considerations for existing cost limits.

* **Tests**
  * Added regression coverage for list and tuple multipliers, non-numeric arguments, and consistent query-cost evaluation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,4 +6,6 @@ For released versions, see the [Releases](https://github.com/mirumee/ariadne/rel
 
 ## Unreleased
 
+### Fixed
 
+- Query cost validation now uses the length of list and tuple arguments named in `multipliers`. Queries that previously passed may exceed `maximum_cost` after upgrading; review affected query costs and thresholds before deployment.
--- a/ariadne/validation/query_cost.py
+++ b/ariadne/validation/query_cost.py
@@ -326,15 +326,15 @@ class CostValidator(ValidationRule):
                 else:
                     val = None
                     break
-            try:
-                if val is not None and not isinstance(val, dict):
+            if isinstance(val, list | tuple):
+                # A list/tuple argument contributes its length, so that e.g.
+                # `field(ids: [ID!]) @cost(multipliers: ["ids"])` costs len(ids).
+                parsed_vals.append(len(val))
+            elif val is not None and not isinstance(val, dict):
+                try:
                     parsed_vals.append(int(val))
-            except (ValueError, TypeError):
-                pass
-        parsed_vals = [
-            len(multiplier) if isinstance(multiplier, list | tuple) else multiplier
-            for multiplier in parsed_vals
-        ]
+                except (ValueError, TypeError):
+                    pass
         return [m for m in parsed_vals if m > 0]
 
     def get_cost_exceeded_error(self) -> GraphQLError:
--- a/docs/06-Extensions/03-query-validators.md
+++ b/docs/06-Extensions/03-query-validators.md
@@ -65,6 +65,18 @@ type Query {
 
 In the above example, final complexity will be multiplied by both `promoted` and `regular` values.
 
+List-valued arguments named in `multipliers` contribute their length:
+
+```graphql
+type Query {
+    posts(ids: [ID!]!): [Post!]! @cost(complexity: 1, multipliers: ["ids"])
+}
+```
+
+Every list argument explicitly named in `multipliers` uses length scaling. To keep an argument from affecting the multiplier, omit its name from `multipliers`; `useMultipliers: false` disables multiplication for the whole field, not for an individual argument. The same behavior applies to Python cost maps, including tuple-valued arguments.
+
+**Upgrade note:** Earlier versions ignored list-valued multipliers. Queries using them can now have higher costs and exceed an existing `maximum_cost` limit. Review representative query costs and thresholds before deploying the update.
+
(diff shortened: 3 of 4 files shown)


jaideeppyne:
Heads up on the red CI: the failing `build` jobs are a pre-existing type-check failure unrelated to this change — `ty` flags `ariadne/schema_visitor.py:720` (a `# ty: ignore[invalid-argument-type]` the newer `ty` no longer accepts), a file this PR doesn't touch. The same failure is currently on `main`. This PR only changes `query_cost.py` and its tests, and `test_query_cost_validation.py` passes (34 tests). Happy to also fix the `schema_visitor.py` type-ignore in a separate PR if that's useful.

jaideeppyne:
Good catch — added `-> None` to both regression tests so the `ty check` recipe passes. Verified locally with `ty check` (clean). The remaining red checks are unrelated infra flakes (a `hatch` download 404 on the fastapi-3.10 integration job and the codecov patch threshold), not this change.

🤖 Addressed with the assistance of Claude Code.

jaideeppyne:
Not a nudge, just flagging why this one shows red, in case it is putting reviewers off.

The failing job is `hatch run types:check`, on an unused `ty: ignore` directive at `ariadne/schema_visitor.py:720`. That file is not touched by this PR, and the same job is failing on `main` (the last few `Tests` runs there are red, going back before this branch existed), so it looks like a `ty` release started reporting the suppression as unfulfilled.

This PR only changes `ariadne/validation/query_cost.py` and its test. Happy to leave it alone, or to drop that stale directive here if you would rather have CI green first.

jaideeppyne:
Addressed the remaining review-body findings in `c077b64`: the newly added regression tests/helper now have scoped parameter and return annotations, and the long DAG test name no longer needs `noqa: E501`. Validation: `ruff check`, `ruff format --check`, and the full `tests/test_query_cost_validation.py` file (`37 passed`).

jaideeppyne:
Addressed the migration-notice feedback in `9a0a06d`. The unreleased changelog now warns that list/tuple multiplier arguments can increase query cost and cause existing queries to exceed `maximum_cost`; the validator documentation includes a list example and asks operators to review costs and thresholds before upgrading.

The docs also clarify the existing controls: omit an argument from `multipliers` to exclude it from scaling, or use `useMultipliers: false` for the whole field. I don't have evidence identifying affected deployments, so I haven't claimed a targeted operator list.

This follow-up changes documentation only. `git diff --check` passes. I attempted `just check`, but it stops because `hatch` is not installed in this checkout's environment; I am not claiming a fresh test-suite pass. Prepared with AI assistance.

